PR #22, opened 2020-12-13
v0.4.0

MaartenGr:
There are several changes and features that I would like to add:
* LDAvis-like visualization
* Option to select language instead of sentence-transformers models to minimize the complexity of using BERTopic
  * Also shows list of languages if incorrect language is selected
  * Uses English embedding model if None is selected
* Improved logging (remove duplicates)
* Update tests 
* Check if BERTopic is fitted 
* Show list of sentence-transformer models if an incorrect model is selected
* More extensive tutorials in the documentation
* Explore the use of TF-IDF instead of transformer models to improve the speed of application

And a few more if I run into them while developing the above.


MaartenGr:
Changed `min_topic_size` from 30 to 15 as a high value results in instable clusters.

MaartenGr:
Numpy for Python 3.6 will be dropped and was therefore removed from the workflow.

alronz:
Hi @MaartenGr 
Very nice additions, thanks for your great work.
I just wanted to mention that [this kind of preprocessing ](https://github.com/MaartenGr/BERTopic/pull/22/files#diff-f6ce44f3a4e62284e53bc5dade21e84ba3471302e18d87bf14e59e8aab99e889R1104) impacts other languages such as Arabic :)

MaartenGr:
@alronz Thanks for the feedback! Would you say that the following code would negatively impact other languages?

```python
cleaned_documents = [doc.lower() for doc in documents]
cleaned_documents = [doc.replace("\n", " ") for doc in cleaned_documents]
cleaned_documents = [doc.replace("\t", " ") for doc in cleaned_documents]
cleaned_documents = [re.sub(r'[^A-Za-z0-9 ]+', '', doc) for doc in cleaned_documents]
cleaned_documents = [doc if doc != "" else "emptydoc" for doc in cleaned_documents]
```

I hoped by only lowering and keeping alphanumeric characters, it would not negatively impact the interpretation of other languages.

alronz:
Yes I tried it out and I guess this part ` [re.sub(r'[^A-Za-z0-9 ]+', '', doc) for doc in cleaned_documents]` removed all Arabic characters

MaartenGr:
I was afraid of that. I might make it easy for me and only apply that specific step if "english" is selected.

alronz:
btw I liked the visualization of topics very much 👍 
however I couldn't use the `visualize_distribution` since doing soft clustering using HDBSCAN `all_points_membership_vectors` took forever (many hours and still not finished for 200k tweets)

MaartenGr:
Thanks! That was something I really wanted to have in the newest version. 
I am a bit confused, did you manage to train a BERTopic instance or did it get stuck at `all_points_membership_vectors`?

alronz:
I had to customize the code to skip `all_points_membership_vectors` and get only topic labels without all the probability distribution stuff.  Clustering is quite fast with the same data size (few minutes). Not sure why doing soft clustering is super slow (I even think it gets stuck since I have to restart the juypter kernal to kill it)

MaartenGr:
Although it should be quite fast, it might be that it gets stuck due to the sheer size of the probabilities that were generated. If the dataset is relatively large, then it is likely that many clusters will be generated and `probabilities` might explode in size. I haven't seen this happening myself though. I could fix this by giving you the option to turn off the extraction of `probabilities`.

alronz:
I think having that option would be really great, thanks @MaartenGr

alronz:
Another minor issues are related to the nature of the Arabic language. For example, Arabic text will be unreadable on the visualizations. It needs to be processed differently. I could even contribute this back if you like.

MaartenGr:
Added a `calculate_probabilities` parameter! 

I intend to publish v0.4 of BERTopic tomorrow. So I do not know if I have the time to properly look at fixing the Arabic language representation on Plotly/Matplotlib as the issue might also extend to other languages. I do want to add it, but it might be after the holidays for a v0.4.1 version.

PR #60, opened 2021-02-17
v0.6

MaartenGr:
The pull request for v0.6. The main feature will be the option to perform dynamic topic modeling (DTM). DTM is a topic modeling technique in which temporal features are taking into account. In other words, it shows the evolution of a topic over time. For example, a topic about environmental awareness may contain different words if you would compare 1995 with 2015.

### Features
* DTM: Added a basic dynamic topic modeling technique based on the global c-TF-IDF representation 
    * `model.topics_over_time(docs, timestamps, global_tuning=True)`
* DTM: Option to evolve topics based on t-1 c-TF-IDF representation which results in evolving topics over time
    * Only uses topics at t-1 and skips evolution if there is a gap
    * `model.topics_over_time(docs, timestamps, evolution_tuning=True)`
* DTM: Function to visualize topics over time 
    * `model.visualize_topics_over_time(topics_over_time)`
* DTM: Add binning of timestamps  
    * `model.topics_over_time(docs, timestamps, nr_bins=10)`
* GENERAL: Add function get general information about topics (id, frequency, name, etc.) 
    *  `get_topic_info()`
* MKDOCS: Add dynamic topic modeling tutorial
    * The plan is currently to use Trump tweets as an example and export the HTML file to mkdocs

### Improvements
* Improved stability of c-TF-IDF by taking the average number of words across all topics instead of the number of documents

### Fixes
*  `_map_probabilities()` does not take into account that there is no probability of the outlier class and the probabilities are mutated instead of copied (#63, #64)

### **DTM Guide**  
To use dynamic topic modeling and create topics over time, a BERTopic instance first needs to be fitted. Then, you can apply `topics_over_time()`:

```python
import re
import pandas as pd
from bertopic import BERTopic

# Prepare data
trump = pd.read_csv('https://drive.google.com/uc?export=download&id=1xRKHaP-QwACMydlDnyFPEaFdtskJuBa6')
trump.text = trump.apply(lambda row: re.sub(r"http\S+", "", row.text).lower(), 1)
trump.text = trump.apply(lambda row: " ".join(filter(lambda x:x[0]!="@", row.text.split())), 1)
trump.text = trump.apply(lambda row: " ".join(re.sub("[^a-zA-Z]+", " ", row.text).split()), 1)
trump = trump.loc[(trump.isRetweet == "f") & (trump.text != ""), :]
timestamps = trump.date.to_list()
tweets = trump.text.to_list()

# Create global topics
model = BERTopic(verbose=True)
topics, _ = model.fit_transform(tweets)

# Topics over time (DTM)
topics_over_time = model.topics_over_time(tweets, topics, timestamps, nr_bins=20)

# Visualize DTM
model.visualize_topics_over_time(topics_over_time, top_n=20)
```


MaartenGr:
> As I experienced, `from_topic` may be -1 too.

Hmmm, that actually should not happen since reducing topics starts from the least frequent topic up to and not including -1... I will have to test that out, thanks!

nadiafelix:
It returned this error: AttributeError: 'BERTopic' object has no attribute 'topics_over_time'

PR #72, opened 2021-03-05
Feature: adding significance score to key word generation

DavidYoo912:
#### Brief description of what is fixed or changed

TF-IDF key word generation: 
added a significance score feature which takes the c-tf-idf matrix and multiplies by each term's n-gram length
This means that longer n-gram terms will score higher and would be more likely to show up as the key word for a topic when running model.get_topic() 

This feature is added as an argument in the BERTopic class so users can flag (True or False) to allow usage of this calculation 

#### Other comments
The code to account for this addition was made in _c_tf_idf() in _bertopic.py
if significance flag is set to True, the c_tf_idf sparse matrix is converted to a numpy array and is multiplied by the array of n-gram for each unique term. 
This result is then converted to a compressed sparse matrix and output as a variable called c_tf_idf


MaartenGr:
Although my response is late, don't think I have forgotten about this one! I am currently testing and evaluating techniques like these to see if they fit within the current API. Depending on the quality of the generated topics I might integrate this into either v0.7 or v0.8. The main issue here is that I do not want to overcomplicate the API with too many hyperparameters. 

I am closing it for now but may re-open it or integrate it into a newer version if I figure out a nice way to keep the hyperparameter-space to a minimum.

PR #74, opened 2021-03-08
Fix hovertemplate indices for LDAviz-like plotting

boscacci:
I had to swap these indices around in order to get "Words: " and "Size: " data to appear in my plots.

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -1471,8 +1471,8 @@ class BERTopic:
 
         # Update hover order
         fig.update_traces(hovertemplate="<br>".join(["<b>Topic %{customdata[2]}</b>",
-                                                     "Words: %{customdata[3]}",
-                                                     "Size: %{customdata[4]}"]))
+                                                     "Words: %{customdata[1]}",
+                                                     "Size: %{customdata[0]}"]))
 
         # Create a slider for topic selection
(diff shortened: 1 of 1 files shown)


MaartenGr:
Strange that you had to change the indices for "Words" and "Size" as I am already getting the correct values in the example Google Colab session. In what environment with which versions are you getting this issue?

boscacci:
I'm on my local (macOS 10.15.6) in a virtualenv, with pyenv Python 3.8.6:
image

Related packages:
image

I have checked to make sure that my jupyter notebook is using the correct/same venv kernel and site packages.

MaartenGr:
@boscacci Which plotly version are you using? 

This will most likely take some testing since we get different results in our environment. The solution you propose could most likely be correct but I cannot be certain until I test it across different environments and package versions.

boscacci:
Sorry for the delay—I'm on Plotly 4.14.3

MaartenGr:
@boscacci Ah, then that issue is most likely caused by using that Plotly version. If you install BERTopic with `pip install bertopic[visualization]` then it will install a Plotly version below 4.14.3 which should fix the issue.

MaartenGr:
Seeing as this issue was most likely caused by an incompatible version of Plotly, I will close this for now. Please re-open this if using the compatible version of Plotly still gives this issue!

boscacci:
I think yes, likely Plotly version issue, thanks

PR #94, opened 2021-04-15
added function to get a coherence score

Viole-Grace:
Following the latest issue : https://github.com/MaartenGr/BERTopic/issues/90

I have added a function to calculate coherence scores in _bertopic.py


bertopic/_bertopic.py
```diff
@@ -984,6 +988,40 @@ def save(self,
...
+
+        cleaned_documents = self._preprocess_text(documents)
+        vectorizer_ = self._vectorizer_model
+        tokenizer_ = vectorizer_.build_tokenizer()
+
+        words = vectorizer_.get_feature_names()
+        tokens = [tokenizer_(doc) for doc in cleaned_documents]
+        dictionary = corpora.Dictionary(tokens)
+        corpus = [dictionary.doc2bow(token) for token in tokens]
+        topic_words_ = [[words for words, _ in self.get_topic(topic)] for topic in range(len(set(topics))-1)]
```
Amine-OMRI, line 1015: topic_words_ = set([[words for words, _ in self.get_topic(topic)] for topic in range(len(set(topics))-1)])
# use a **set** instead of a **list** to obtain unique words that will improve the NPMI score

bertopic/_bertopic.py
```diff
@@ -984,6 +988,40 @@ def save(self,
...
+              * Clean documents using inbuilt preprocessing
+              * Initialize inbuilt CountVectorizer
+              * Build a dictionary using tokens from the tokenized clean documents
+              * Get all topic words from all topics, build a coherence model
+
+
+        """
+        
+        if coherence == None:
+            coherence = 'c_v'
```
Amine-OMRI, line 1005: def get_coherence_score(self, documents : List[str], coherence='c_v') -> float:
setting the defaul coherence to 'c_v'

Viole-Grace: CoherenceModel uses gensim. Should I include gensim in the requirements?

PR #95, opened 2021-04-15
fixed changes and added function to get coherence score

Viole-Grace:
- Fixed coding style changes to optimize/improve readability
- Added function to get a coherence score (from the original pull request)


MaartenGr:
Thank you for the pull request. There are two things that make me hesitant to implement this pull request. First, the additional requirement. If Gensim is not needed for the algorithm itself, then adding this requirement will dilute the requirement space which potentially could lead to dependency issues. 

Second, the addition of an evaluation technique that has significant flaws. By adding topic coherence to BERTopic I communicate to the user that it is an appropriate measure for evaluating the quality of the generated topic model. Don't get me wrong, it is not by definition a poor measurement, but it also should not be used by itself as the ground truth for evaluation.

Viole-Grace:
Hey!

I understand. I have been studying your method and was working around
building my own topic modeling method, where I replaced UMAP with an
autoencoder and wanted to check for fuzzy clusters instead. Since I work in
R&D, my peers needed me to benchmark it. Given topic modeling as a method,
since it is such a subjective and ambiguous matter, only way of quantifying
it to someone who does not understand it is giving it a metric, or a score.

Using it for anything other than academia, we need to quantify it with a
score to make it usable -- as far as I understand.

Totally agree about the gensim though. Would it be better if I wrote the
source code particularly for coherence calculation instead?

Thanks for the reply though!

- Aamir

MaartenGr:
> Given topic modeling as a method, since it is such a subjective and ambiguous matter, only way of quantifying
> it to someone who does not understand it is giving it a metric, or a score.

To me, this is actually something you should avoid as much as possible. Since it is such a subjective matter, returning an imperfect evaluation measure actually communicates to the user that this evaluation measure is without any issues. The result will be that users are going to optimize BERTopic based on coherence scores such as `c_v` since it would be implemented in the model, while in practice this is very much not advised.

PR #113, opened 2021-04-30
Fix typo

dplutcho:
This is my first time doing this so hopefully I am doing this appropriately.


MaartenGr:
Thank you for spotting the typo. I'll make sure to fix this in #111 since a typo is a bit too small to be added through a pull request.

PR #117, opened 2021-05-07
fix small spelling error.

dplutcho:


MaartenGr:
Thank you for spotting the typo. I'll make sure to fix this in #111 since a typo is a bit too small to be added through a pull request.

PR #136, opened 2021-06-04
Update SentenceTransformers model

nreimers:
Hi,
I love the nice repository you have. Great work!

Over the last weeks, I trained substantially better sentence-transformers models which are now available. The two models I can recommend are:
paraphrase-mpnet-base-v2 <= for English
paraphrase-multilingual-mpnet-base-v2   <= for 50+ languages

The  models are substantially better than any previous model. For example, when evaluating for clustering on the 20Newsgroups dataset, the performance (V-Measure) of the paraphrase-mpnet-base-v2 is at 47.8 while the currently used model in BERTTopic (distilbert-base-nli-mean-tokens) is just at 27.9 => We have a +20 points improvement.

The new models have been extensively evaluated on 15 different datasets and task, and they are substantially better on all datasets. The largest gains are for noisy data like user written text (Twitter, Reddit etc.)

If you like faster models, I can recommend:
paraphrase-MiniLM-L6-v2    <= for English
paraphrase-multilingual-MiniLM-L12-v2  <= for 50+ languages

========================

In this pull request, I hope I replaced all references of the old model to the new paraphrase-mpnet-base-v2 model. Also I removed at two places the manual setting of the device. The device is automatically determined, so usually it is better not to set it manually.

Let me know if you  have further questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ You can select any model from sentence-transformers [here](https://www.sbert.net
 and pass it to BERTopic:
 
 ```python
-topic_model = BERTopic(embedding_model="paraphrase-mpnet-base-v2")
+topic_model = BERTopic(embedding_model="paraphrase-MiniLM-L6-v22")
 ```
 
 Or select a SentenceTransformer model with your own parameters:
@@ -139,7 +139,7 @@ Or select a SentenceTransformer model with your own parameters:
 ```python
 from sentence_transformers import SentenceTransformer
 
-sentence_model = SentenceTransformer("paraphrase-mpnet-base-v2")
+sentence_model = SentenceTransformer("paraphrase-MiniLM-L6-v2")
 topic_model = BERTopic(embedding_model=sentence_model)
 ```
 
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ You can select any model from sentence-transformers [here](https://www.sbert.net
 and pass it to BERTopic:
 
 ```python
-topic_model = BERTopic(embedding_model="xlm-r-bert-base-nli-stsb-mean-tokens")
+topic_model = BERTopic(embedding_model="paraphrase-mpnet-base-v2")
 ```
 
 Or select a SentenceTransformer model with your own parameters:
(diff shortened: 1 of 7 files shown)


MaartenGr:
First, thank you for training the new models! Second, You beat me to it! I have been experimenting with the two new sets of models and the results have been very promising with respect to both the topic clustering as well as the topic representations. They seem much more coherent now regardless of whether I use `paraphrase-MiniLM-L6-v2` or `paraphrase-mpnet-base-v2`. The multi-lingual models also demonstrated much better performance on Dutch texts than the previous ones. 

Quick question though, I am actually leaning towards the `MiniLM` models to use as a default due to its balance between speed and performance. Moreover, especially within topic modeling, speed seems to be a significant barrier for many users. Seeing as you actually trained the models, would you suggest one over the other in general use?

nreimers:
Hi @MaartenGr 
Happy to hear that :)

Yes, the paraphrase-MiniLM-L6-v2 also produces really good results in my experiments on clustering Reddit data. I often use that model due to the better speed and the subjective feeling about the quality is comparable to the mpnet models.
 
So when speed is a barrier for the users, using `paraphrase-MiniLM-L6-v2`  and `paraphrase-multilingual-MiniLM-L12-v2` as the default can make sense and then pointing them to the mpnet-models if they want higher quality.

nreimers:
Updated the PR so that the MiniLM models are used. Added a note in the FAQ that paraphrase-mpnet-base-v2  and paraphrase-multilingual-mpnet-base-v2 can be used if a higher quality is desired.

MaartenGr:
Great, thanks for the changes. I am currently running the workflows in Github but it seems that it encounters TimeOutErrors quite frequently when trying to access sbert.net and the corresponding models. Perhaps a version update of sentence-transformers might be a solution or extending the timeout limit in the Github workflow. 

Either way, I intend to merge this today or tomorrow when I tested some things out regardless of whether it passes the workflow since it works fine in various other environments.

zhongpeixiang:
Regarding the best model, according to the benchmark result in https://www.sbert.net/docs/pretrained_models.html, the model ```all-MiniLM-L6-v2``` has better performance but comparable speed than ```paraphrase-MiniLM-L6-v2```, which one would you recommend?

nreimers:
The all-* models: Better and same speed as the paraphrase-MiniLM-L6-v2 model

The all-* models were recently released.

zhongpeixiang:
> The all-* models: Better and same speed as the paraphrase-MiniLM-L6-v2 model
> 
> The all-* models were recently released.

I see, then i will use the recent all-* models. Thank you for the quick reply.

PR #153, opened 2021-06-23
Fix issue with UMAP import

xoubinha:
- Modify UMAP import.

Explanation:
When trying to import `bertopic` I was getting an AttributeError related to UMAP import, saying:
`module umap has no attribute UMAP`
I checked that I had the correct version of `umap-learn` but still it wasn't working. Changing the imports as in the PR solved the issue. 😊

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -13,7 +13,7 @@ from typing import List, Tuple, Union, Mapping, Any
 
 # Models
 import hdbscan
-from umap import UMAP
+import umap.umap_ as UMAP
 from sklearn.feature_extraction.text import CountVectorizer
 from sklearn.metrics.pairwise import cosine_similarity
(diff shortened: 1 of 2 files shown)


MaartenGr:
The changes that you propose are workarounds to the issues mentioned [here](https://github.com/lmcinnes/umap/issues/24) and [here](https://github.com/lmcinnes/umap/issues/534). To understand whether the change is truly necessary, it is important that we understand why you are facing this issue. Did you have a previous version of `umap` installed together with `umap-learn`? That typically seems to be the issue as many install UMAP with `pip install umap` instead of `pip install umap-learn`. Also, did you try it out in a fresh environment?

xoubinha:
I did it in a fresh environment, but it must have used already installed packages. I tried it again uninstalling everything and it did work.

PR #176, opened 2021-07-10
v0.9

MaartenGr:
## Highlights
* Get the most representative documents per topic: `topic_model.get_representative_docs(topic=1)`
    * This allows users to see which documents are good representations of a topic and better understand the topics that were created
* Added `normalize_frequency` parameter to `visualize_topics_per_class` and `visualize_topics_over_time` in order to better compare the relative topic frequencies between topics
* Return flat probabilities as default, only calculate the probabilities of all topics per document if `calculate_probabilities` is True
* Implemented a guided BERTopic by defining seed topics:

```python
# NOTE: The Reuters dataset was ued
seed_topic_list = [["company", "billion", "quarter", "shrs", "earnings"],
                   ["acquisition", "procurement", "merge"],
                   ["exchange", "currency", "trading", "rate", "euro"],
                   ["grain", "wheat", "corn"],
                   ["coffee", "cocoa"],
                   ["natural", "gas", "oil", "fuel", "products", "petrol"]]

topic_model = BERTopic(seed_topic_list=seed_topic_list)
topics, _ = topic_model.fit_transform(docs)
```

Guided BERTopic works in two ways. 

First, we create embeddings for each seeded topics by joining them and passing them through the document embedder. These embeddings will be compared with the existing document embeddings through cosine similarity and assigned a label. If the document is most similar to a seeded topic, then it will get that topic's label. If it is most similar to the average document embedding, it will get the -1 label. These labels are then passed through UMAP to create a semi-supervised approach that should nudge the topic creation to the seeded topics. 

Second, we take all words in `seed_topic_list` and assign them a multiplier larger than 1. Those multipliers will be used to increase the IDF values of the words across all topics thereby increasing the likelihood that a seeded topic word will appear in a topic. This does, however, also increase the chance of an inrelevant topic having unrelated words. In practice, this should not be an issue since the IDF value is likely to remain low regardless of the multiplier. The multiplier is now a fixed value but may change to something more elegant, like taking the distribution of IDF values and its position into account when defining the multiplier. 

## Fixes
* Fix loading pre-trained BERTopic model
* Fix mapping of probabilities

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ from sklearn.datasets import fetch_20newsgroups
 docs = fetch_20newsgroups(subset='all',  remove=('headers', 'footers', 'quotes'))['data']
 
 topic_model = BERTopic()
-topics, _ = topic_model.fit_transform(docs)
+topics, probs = topic_model.fit_transform(docs)
 ```
 
 After generating topics, we can access the frequent topics that were generated:
@@ -161,7 +161,7 @@ You can also use previously generated embeddings by passing it to `fit_transform
 
 ```python
 topic_model = BERTopic()
-topics, _ = topic_model.fit_transform(docs, embeddings)
+topics, probs = topic_model.fit_transform(docs, embeddings)
 ```
 
 ## Dynamic Topic Modeling
@@ -186,7 +186,7 @@ Then, we need to extract the global topic representations by simply creating and
 
 ```python
 topic_model = BERTopic(verbose=True)
-topics, _ = topic_model.fit_transform(tweets)
+topics, probs = topic_model.fit_transform(tweets)
 ```
 
 From these topics, we are going to generate the topic representations at each timestamp for each topic. We do this 
--- a/README.md
+++ b/README.md
@@ -10,8 +10,12 @@
 <img src="images/logo.png" width="35%" height="35%" align="right" />
 
 BERTopic is a topic modeling technique that leverages 🤗 transformers and c-TF-IDF to create dense clusters
-allowing for easily interpretable topics whilst keeping important words in the topic descriptions. BERTopic supports 
-guided, (semi-)supervised, and dynamic topic modeling. It even supports visualizations similar to LDAvis! 
+allowing for easily interpretable topics whilst keeping important words in the topic descriptions.
+
+BERTopic supports 
+[**guided**](https://maartengr.github.io/BERTopic/tutorial/guided/guided.html), 
+(semi-) [**supervised**](https://maartengr.github.io/BERTopic/tutorial/supervised/supervised.html), 
+and [**dynamic**](https://maartengr.github.io/BERTopic/tutorial/topicsovertime/topicsovertime.html) topic modeling. It even supports visualizations similar to LDAvis!
 
 Corresponding medium posts can be found [here](https://towardsdatascience.com/topic-modeling-with-bert-779f7db187e6?source=friends_link&sk=0b5a470c006d1842ad4c8a3057063a99) 
 and [here](https://towardsdatascience.com/interactive-topic-modeling-with-bertopic-1ea55e7d73d8?sk=03c2168e9e74b6bda2a1f3ed953427e4).
@@ -114,7 +118,6 @@ topic_model.visualize_topics()
 
 <img src="images/topic_visualization.gif" width="60%" height="60%" align="center" />
 
-
 ## Embedding Models
 BERTopic supports many embedding models that can be used to embed the documents and words:
 * Sentence-Transformers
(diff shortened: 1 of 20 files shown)


Akhamis01 (changes requested):
Hello!

Just wanted to note that this update to improve the probabilities and fix the mappings has introduced a new bug. When using the Bertopic() function with the 'nr_topics' parameter, it will return an error.

To reproduce this error, run:
topic_model = BERTopic(nr_topics=10)

This will return an error (Key error: 10)

MaartenGr:
@Akhamis01 Nice catch! Indeed, there are still some bugs regarding this to figure out. There is a fix coming though that should remedy this issue.

Akhamis01:
Sounds good! Also, just a side note, have you considered creating a function that would return the optimal number of topics to be used within a topic model? I feel that would be extremely useful :)

MaartenGr:
@Akhamis01 You can already get the optimal number of topics by setting `nr_topics = "auto"` when instantiating BERTopic.

Akhamis01:
> @Akhamis01 You can already get the optimal number of topics by setting `nr_topics = "auto"` when instantiating BERTopic.

Thank you! Just wondering, how is the optimal number of topics calculated for a topic model? Is it through Coherence scores?

MaartenGr:
@Akhamis01 The reduction is based on the similarity between topics. If two topics are very similar to each other, they will be merged. Coherence scores have a bunch of issues that make it difficult to properly evaluate a topic model using only one method.

Akhamis01:
> @Akhamis01 The reduction is based on the similarity between topics. If two topics are very similar to each other, they will be merged. Coherence scores have a bunch of issues that make it difficult to properly evaluate a topic model using only one method. 

Alright, and how is the similarity between the topics calculated in general? What metrics are the similarities based on?

MaartenGr:
@Akhamis01 HDBSCAN is used to cluster the topics even further based on their topic vectors. You can find more about that here:

https://github.com/MaartenGr/BERTopic/blob/687d84612a62c5eeddcc089a4e29cbb4710ba23e/bertopic/_bertopic.py#L1568

If you have any further questions, could you create a separate issue for that? It helps keeping discussions relevant to this specific pull request.

Akhamis01:
Makes sense. Thank you for your help, appreciate it! I'll create a new issue for any other questions :)

farshi:
@MaartenGr GuidedBERTopic idea and implementations looks great!  I will give it a try and let you know my feedback. Thank you

FedePacio97:
> ## Highlights
> * Get the most representative documents per topic: `topic_model.get_representative_docs(topic=1)`
>   
>   * This allows users to see which documents are good representations of a topic and better understand the topics that were created
> * Added `normalize_frequency` parameter to `visualize_topics_per_class` and `visualize_topics_over_time` in order to better compare the relative topic frequencies between topics
> * Return flat probabilities as default, only calculate the probabilities of all topics per document if `calculate_probabilities` is True
> * Implemented a guided BERTopic by defining seed topics:
> 
> ```python
> # NOTE: The Reuters dataset was ued
> seed_topic_list = [["company", "billion", "quarter", "shrs", "earnings"],
>                    ["acquisition", "procurement", "merge"],
>                    ["exchange", "currency", "trading", "rate", "euro"],
>                    ["grain", "wheat", "corn"],
>                    ["coffee", "cocoa"],
>                    ["natural", "gas", "oil", "fuel", "products", "petrol"]]
> 
> topic_model = BERTopic(seed_topic_list=seed_topic_list)
> topics, _ = topic_model.fit_transform(docs)
> ```
> 
> Guided BERTopic works in two ways.
> 
> First, we create embeddings for each seeded topics by joining them and passing them through the document embedder. These embeddings will be compared with the existing document embeddings through cosine similarity and assigned a label. If the document is most similar to a seeded topic, then it will get that topic's label. If it is most similar to the average document embedding, it will get the -1 label. These labels are then passed through UMAP to create a semi-supervised approach that should nudge the topic creation to the seeded topics.
> 
> Second, we take all words in `seed_topic_list` and assign them a multiplier larger than 1. Those multipliers will be used to increase the IDF values of the words across all topics thereby increasing the likelihood that a seeded topic word will appear in a topic. This does, however, also increase the chance of an inrelevant topic having unrelated words. In practice, this should not be an issue since the IDF value is likely to remain low regardless of the multiplier. The multiplier is now a fixed value but may change to something more elegant, like taking the distribution of IDF values and its position into account when defining the multiplier.
> 
> ## Fixes
> * Fix loading pre-trained BERTopic model
> * Fix mapping of probabilities

Hi, thanks for the update.
Can we decide how many documents we would like to get from this method? For instance, instead of 3 documents, could we get X number of documents. As an example, a kind of method like this: topic_model.get_representative_docs(topic=1, nr_of_documents_returned=20).

MaartenGr:
@FedePacio97 That is currently not possible since only three documents per topic are saved in order to prevent the topic model from saving too many documents. Moreover, this value would then depend on the `min_topic_size` as well as the underlying algorithm that is being used, since this method is only used for HDBSCAN. 

Instead, you can use the `topics` output to see which documents belong to which topic. Also, you could use [`.visualize_documents`](https://maartengr.github.io/BERTopic/getting_started/visualization/visualization.html#visualize-documents) to get an understanding of the documents that occupy each topic and which might be best representative.

FedePacio97:
Thank you very much for the fast reply.
This is definitely what I was looking for.

PR #179, opened 2021-07-15
Plotting minor fixes: use consistent plotly template etc.

davidtweddell:
Hello Maarten. This is a great package, and I really like the addition of the visualizations.

Looking at the results and the source, I spotted a few inconsistencies in how plots are styled. My preference is for consistent plots, so I forked and: 
- edited files in `plotting/` to use template to `plotly_white` if they weren't already using it 
- made couple of other minor changes in `_topics_over_time.py` to comment out custom colours (so that colours on barcharts and timeseries plot match)
- removed specificed background colour in `_hierarchy.py` so that template background is applied

If, in fact, these were all intentional design decisions, then no worries - you can ignore this pull request.

Thanks.

David

### DIFF
--- a/bertopic/plotting/_distribution.py
+++ b/bertopic/plotting/_distribution.py
@@ -72,6 +72,7 @@ def visualize_distribution(topic_model,
     )
 
     fig.update_layout(
+        template="plotly_white",
         xaxis_title="Probability",
         title={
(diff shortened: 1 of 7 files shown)


MaartenGr:
Thank you for the pull request, most of the ones you mentioned were by design. However, I'll see if some might be relevant for the next release.

PR #200, opened 2021-08-11
Fix embeddings tutorial typo at spaCy section

vmussa:
Just a small fix at the embeddings tutorial at the spaCy section. It seems that this piece of the Flair instructions had been left at the spaCy's one:

``` 
allows you to choose almost any embedding model that is publicly available. Flair can be used as follows:
```


MaartenGr:
Thank you for the fix! It will soon be implemented in the [next release](https://github.com/MaartenGr/BERTopic/pull/239).

PR #233, opened 2021-09-10
Fix link in Readme.md

louisguitton:


MaartenGr:
Thank you for the fix! It will soon be implemented in the [next release](https://github.com/MaartenGr/BERTopic/pull/239).

PR #318, opened 2021-11-09
Add option to compress saved model

simonfelding:
joblib supports compression and I could really use this feature in my project. lzma and zlib are in the python standard library so there's no added dependencies. Joblib supports more compression algorithms but there really isn't much point in suggesting more compression algorithms.

Joblib automatically detects compressed files when loading, so there won't be any backwards compatibility issues either.

[see this link for examples](https://joblib.readthedocs.io/en/latest/auto_examples/compressors_comparison.html). I could reduce the size of my saved model by 50% using lzma!

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -1227,11 +1227,13 @@ class BERTopic:
 
     def save(self,
              path: str,
+             compression: str = None,
              save_embedding_model: bool = True) -> None:
         """ Saves the model to the specified path
 
         Arguments:
             path: the location and name of the file you want to save
+            compression: Compress the file with "lzma" (slow, very efficient) or "zlib" (fast, efficient). Defaults to None.
             save_embedding_model: Whether to save the embedding model in this class
                                   as you might have selected a local model or one that
                                   is downloaded automatically from the cloud.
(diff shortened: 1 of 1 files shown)


MaartenGr:
Sorry for the late response and thank you for the pull request. It will take some time to test this out across various environments so I cannot make any promises it will be added in the next release. However, I'll put it on the to-do list!

simonfelding:
No worries! Thanks a lot for the work you do on this project :)

I don't see how this would introduce any regressions in any environments, [it's been a feature in joblib since 2018](https://github.com/joblib/joblib/commits/master/joblib/compressor.py). The only potential issue I see is that lzma compression wasn't built-in to python before v3.3.

MaartenGr:
Fortunately, that is not necessarily what I am worried about. The thing is, by opening up the `compress` variable in joblib, there is actually quite a bit more that you open up than "lzma" or "zlib" as can be seen [here](https://joblib.readthedocs.io/en/latest/generated/joblib.dump.html). The compression level and many other compressors (e.g., ‘zlib’, ‘gzip’, ‘bz2’, ‘lzma’ ‘xz’) are being given access to and at this stage, I cannot promise any issues stemming from any combination of compressor and compression level. Having said that, if you output the model as `model.lzma` it will automatically detect that compressor and make use of it.

simonfelding:
> The compression level and many other compressors (e.g., ‘zlib’, ‘gzip’, ‘bz2’, ‘lzma’ ‘xz’) are being given access to and at this stage, I cannot promise any issues stemming from any combination of compressor and compression level. 

These are all ancient compression methods that work flawlessly, battle tested and used for all backups since the 90's :). Yes they are actually all exposed, but there's no reason to choose any of the others for compressing binary data. The only issue a high compression level makes is slowing down saving and loading a bit. You will not lose data by saving a compressed file using these algorithms. It's like uploading something to github, it uses zlib compression and some custom stuff.

> Having said that, if you output the model as model.lzma it will automatically detect that compressor and make use of it.

Yes that would also be a fine way to specify it (although a bit confusing), except you can't specify compression and the default in joblib is 0, meaning no compression.

MaartenGr:
Closing this as the compressor can be selected using the filename.

Issue #333, opened 2021-11-17
numpy >=1.20.0 doesn't support Python3.6 causing BERTopic installation failure


electrowizard:
Hi Maarten,

Thanks for maintaining this amazing project!

I recently tried to upgrade to the latest version of BERTopic from [v0.6.0](https://github.com/MaartenGr/BERTopic/releases/tag/v0.6.0) while using Python 3.6. The installation failed with the error `ERROR: No matching distribution found for numpy>=1.20.0`

On digging deeper, I noticed that even though BERTopic claims support for [Python 3.6](https://github.com/MaartenGr/BERTopic/blob/15ea0cd804d35c1f11c6692f33c3666b648dd6c8/setup.py#L81), numpy versions 1.20.0 and above has depreciated support for Python 3.6. Since the latest version of BERTopic required numpy>=1.20.0, the installation failed.

From the [numpy 1.20.0 release notes](https://numpy.org/devdocs/release/1.20.0-notes.html):

>The Python versions supported for this release are 3.7-3.9, support for Python 3.6 has been dropped

It might be worth re-visiting the minimum supported Python version for BERTopic.

Hope it helps!
Sathvik

MaartenGr:
Nice catch! It indeed does not support python 3.6 anymore and it seems that I did not properly update the readme. Thank you for noticing the inconsistency! I'll make sure to fix this in the next release 😃

doubianimehdi:
I'm not sure what the implications of that are ... does that mean that we can no longer have the latest bertopic version on python 3.6 ? if so what is the latest version we can have on 3.6 ? I'm struggling to install it on dataiku maybe because of that i'm stuck on 0.6.0 ...

MaartenGr:
Typically, this should not affect that many applications as most support at least python 3.7. [Dataiku also supports 3.7](https://doc.dataiku.com/dss/latest/code-envs/operations-python.html) so it should be no problem using the newest version of BERTopic there. If, however, you need to have 3.6 for some reason, you could try an earlier version of BERTopic but many features will be missing. 

Which python version do you have on dataiku? Which issues are you running into?

doubianimehdi:
I have a 3.6 python environnement and i'm trying to have the topics over time and other visualization graphics to run on dataiku so I can make a dashboard out of it and i have 3 weeks left to do that ....

MaartenGr:
[...]

Issue #330, opened 2021-11-16
Customize similarity of topics reduction


paulthemagno:
Hi! I saw there is this parameter: `nr_topics` with this description:

```
Specifying the number of topics will reduce the initial number of topics to the value specified.
This reduction can take a while as each reduction in topics (-1) activates a c-TF-IDF calculation.
If this is set to None, no reduction is applied. Use "auto" to automatically reduce topics that have a similarity of at least 0.9, do not maps all others.
```

How can I change the percentage of similarity? Example: merge together similar topics at 0.8.

I see that for example with `nr_topics= "auto"` the class calls this function: [_auto_reduce_topics](https://github.com/MaartenGr/BERTopic/blob/15ea0cd804d35c1f11c6692f33c3666b648dd6c8/bertopic/_bertopic.py#L1727) but I don't see where `0.9` is put as percentage of similarity.

Thank you!

MaartenGr:
Good catch! It seems that the description is currently outdated as the "auto" step does not use that similarity value anymore. Whenever you use `nr_topics="auto"`, a fresh HDBSCAN instantiation is used to cluster all topics automatically. After the topics are clustered, new topic representations are generated. 

I have put this on the to-do list and will most likely be fixed in the new release!

paulthemagno:
Thanks, @MaartenGr. In your opinion, at the moment, which are the best parameters to pass to bertopic to have the best topics reduction?

MaartenGr:
That very much depends on what you define as "best topics". If, for example, you know there are a certain amount of topics in the documents and not much more than that, setting `nr_topics` to a specific integer would be preferred. However, setting it to "auto" does cluster the topics more naturally. I cannot say which one is better than the other but starting out with "auto" typically gives you a good understanding of the number topics, in their reduced state, in the documents.

MaartenGr:
This was fixed in the v0.9.4 release of BERTopic, so I'll be closing this issue for now. Let me know if you run into any other issues!

Issue #88, opened 2021-03-29
No topic words in visualization


sanshine2020:
Hi Maarten,

I'm calling the BERTopic visualization, but don't see the topic words nor size in the plot. Instead the hoverdata shows customdata[3] and customdata[4], see screenshot. The topic (customdata[2]) however is shown correctly. This issue occurs with the 20 newsgroups dataset as well as with my own data (list of strings). Do you know how to solve this issue?

![image](https://user-images.githubusercontent.com/61740021/112806423-0be50b00-9077-11eb-8cf4-e945c2c0f974.png)

data = fetch_20newsgroups(subset='test', remove=('headers', 'footers', 'quotes'))['data']
topic_model = bertopic.BERTopic()
labels, probabilities = topic_model.fit_transform(data)
fig = topic_model.visualize_topics()
fig.show()

Thnx in advance!

MaartenGr:
Hmmm, I think I have seen this issue before. So far, this has often been caused by an incorrect version of Plotly. I believe you can fix this by installing BERTopic with: `pip install bertopic[visualization]`

Which version of Plotly are you currently using?

juliechoong:
> 
> 
> Hmmm, I think I have seen this issue before. So far, this has often been caused by an incorrect version of Plotly. I believe you can fix this by installing BERTopic with: `pip install bertopic[visualization]`
> 
> Which version of Plotly are you currently using?

I faced the same issue too. This works!

MaartenGr:
Great! Then I will close this for now. The next update will include plotly into the base packages so these issues are unlikely to appear in the future.

stepwei:
Hi Maarten, this error indeed still persists with plotly 5.4.0

MaartenGr:
The dependencies with respect to plotly are currently limited to 4.14.2 as newer versions require a different way of showing words and size. Unfortunately, I do not think I am able to support all versions. In practice, that would mean either `plotly>=4.7.0,<4.14.3` or `plotly>=4.14.3` would be supported. I might opt for the latter if there is enough support to do so and if this does not interfere with common packages.

PR #335, opened 2021-11-18
v0.9.4

MaartenGr:
A number of fixes, documentation updates, and small features:

* Expose diversity parameter
   * Use `diversity=0.1` to change how diverse the words in a topic representation are (ranges from 0 to 1)
* Improve stability of topic reduction by only computing the cosine similarity within c-TF-IDF and not the topic embeddings
   * This also means that the topic embeddings could be improved or do not work properly with certain language models
   * Perhaps this could be improved by averaging out the exemplars but that still would not fully represent a topic
   * c-TF-IDF seems to hold the most information with respect to the topics
* Drop python 3.6
   * #333
* Major documentation overhaul (mkdocs, tutorials, FAQ, images, etc. )
   * #330
* Add property to c-TF-IDF that all IDF values should be positive (has no influence on performance!)
   * #351
   * Note that this should have no influence on performance, it just adds a nice property to have 🙂
* Relax plotly dependency
   * #88 
* Additional logging for `.transform`
   *  #356


eyang9002:
After dropping Python 3.7 support, do you suggest we stay on 3.6? Or > 3.7 moving forward? I've had to do a lot of work to maintain my dev environments and retraining of models using updated versions of BERTopic to get the newest enhancements.

MaartenGr:
I would suggest going with a more recent version of python, such as 3.9. The thing is, security updates for numpy have stopped for python 3.6 so it was necessary to drop that version as it was not supported anymore.

PR #342, opened 2021-11-24
Added support for barplotting more than 9 topics

simonfelding:
Before this patch, the subplots became smaller and smaller the more topics you added. Was unusable above 18 topics and **_broke entirely_** when slightly larger than that, and with cryptic errors too. Width and height controls were exposed but required very careful and frustrating study of how each added and removed row would change the size of each subplot.

Now the subplots are more or less the same size all the time. Instead of height and width settings it now takes number of desired columns instead and scales size according to number of columns and number of topics. Changed default columns from 3 to 4 because it's a bit less wasteful. 3 columns still looks nice though. I tested it thoroughly within the confines of the data I'm working with and it looks good to go.

The reason for the strange-looking `height=200+25*top_n_topics` is that when top_n_topics are sufficiently small, the plots became very small without a minimum size. 200 is a good minimum height in case of very few topics. I can't really explain why the horizontal and vertical spacing works the way it does now, but it really does work well. It has something to do with how plotly automatically scales everything in a confusing way.

I changed the positioning of the title slightly because it decided to place itself inside the plots when plotting about 200 topics.

I removed the x-axis labeling because it was misleading because it's per column. That means a 20px long bar in column 1 means something entirely different than a 20px long bar in column 2 and so on. It also isn't really that meaningful to look at actual probabilities like that anyway, what matters is the proportion between the bars in my opinion. Maybe it would be cool to normalize the length of the bars per subplot, so the total adds up to 1? I think that would make sense.

I spent quite a while tweaking the scaling, but a perfectionist might be able to make it slightly better by actually measuring pixels per row instead of estimating it empirically like I did. At least it's usable and working as intended now, and looking good too.

It would be nice to scale horizontal spacing and width by the longest label, because currently it's possible to overlap with extremely long labels, and shortening them doesn't really make a lot of sense to me.

### DIFF
--- a/bertopic/plotting/_barchart.py
+++ b/bertopic/plotting/_barchart.py
@@ -7,10 +7,9 @@ from plotly.subplots import make_subplots
 
 def visualize_barchart(topic_model,
                        topics: List[int] = None,
-                       top_n_topics: int = 6,
+                       top_n_topics: int = 8,
                        n_words: int = 5,
-                       width: int = 800,
-                       height: int = 600) -> go.Figure:
+                       columns: int = 4) -> go.Figure:
     """ Visualize a barchart of selected topics
 
     Arguments:
@@ -18,8 +17,7 @@ def visualize_barchart(topic_model,
         topics: A selection of topics to visualize.
         top_n_topics: Only select the top n most frequent topics.
         n_words: Number of words to show in a topic
-        width: The width of the figure.
-        height: The height of the figure.
+        columns: Number of columns to plot.
 
     Returns:
         fig: A plotly figure
@@ -52,13 +50,13 @@ def visualize_barchart(topic_model,
 
     # Initialize figure
     subplot_titles = [f"Topic {topic}" for topic in topics]
-    columns = 3
+    columns = 4
     rows = int(np.ceil(len(topics) / columns))
     fig = make_subplots(rows=rows,
                         cols=columns,
                         shared_xaxes=True,
                         horizontal_spacing=.15,
-                        vertical_spacing=.15,
+                        vertical_spacing=.2/rows,
                         subplot_titles=subplot_titles)
 
     # Add barchart for each topic
@@ -86,24 +84,24 @@ def visualize_barchart(topic_model,
         showlegend=False,
         title={
             'text': "<b>Topic Word Scores",
-            'y': .95,
-            'x': .15,
-            'xanchor': 'center',
-            'yanchor': 'top',
             'font': dict(
                 size=22,
                 color="Black")
         },
-        width=width,
-        height=height,
+        width=220*columns,
+        autosize=False,
+        height=200+25*top_n_topics,
+        uniformtext={"mode": "show", "minsize": 8},
+        margin_t=100,
         hoverlabel=dict(
             bgcolor="white",
             font_size=16,
             font_family="Rockwell"
         ),
     )
-
-    fig.update_xaxes(showgrid=True)
-    fig.update_yaxes(showgrid=True)
+    
+    fig.update_xaxes(showgrid=True,automargin=True,showticklabels=False)
+    fig.update_yaxes(showgrid=True,automargin=True,nticks=n_words)
 
     return fig
+  
--- a/bertopic/plotting/_barchart.py
+++ b/bertopic/plotting/_barchart.py
@@ -9,6 +9,8 @@ def visualize_barchart(topic_model,
                        topics: List[int] = None,
                        top_n_topics: int = 8,
                        n_words: int = 5,
+                       title: str = 'Topic Word Scores',
+                       normalize: bool = True,
                        columns: int = 4) -> go.Figure:
     """ Visualize a barchart of selected topics
 
@@ -18,6 +20,8 @@ def visualize_barchart(topic_model,
         top_n_topics: Only select the top n most frequent topics.
         n_words: Number of words to show in a topic
         columns: Number of columns to plot.
+        normalize: Sets x-axis max value to highest score (True, default) or 1 (False).
+        title: Sets title of plot.
 
     Returns:
         fig: A plotly figure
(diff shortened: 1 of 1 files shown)


MaartenGr:
Thank you for the extensive description and pr! 

There are a few things though that I am not entirely sure of. First, the removal of `width` and `height`. The size of the figure is something I want to expose regardless of any automatic scaling to make sure users get direct customization of quite an important parameter set. 

Second, the different open pull requests that you made are focused on changing the same visualization structure so it would make sense to combine them and apply the changes to every single visualization plot. Otherwise, things might get a bit muddled. 

Third, these changes will actually break the [visualization code](https://github.com/MaartenGr/BERTopic/blob/15ea0cd804d35c1f11c6692f33c3666b648dd6c8/bertopic/_bertopic.py#L873) within BERTopic as the `plotting` functions are directly referenced within the main `_bertopic.py` code. So any changes in the plotting itself, including docstrings and pointers to parameters, will also need to be changed in `_bertopic.py`. 

Having said that, I am currently working on v0.9.4 that will include fixes and some additional features. In this new version, I can look into making the automatic scaling available across all plotting functions whilst keeping the sizing parameters and perhaps let them inherit from an abstract plotting class to make sure all things are equal in basis across all plotting functions.

simonfelding:
> There are a few things though that I am not entirely sure of. First, the removal of width and height. The size of the figure is something I want to expose regardless of any automatic scaling to make sure users get direct customization of quite an important parameter set.
Well you could let the user use the width and height controls, but I think it would make more sense to use my scaling logic and suggest the user exports to svg and rescale that instead. Otherwise you're going to have a situation where the text doesn't scale, and the proportions get messed up, and so on. I really don't think the width and height is useful at all to expose except to compensate for unexpected behavior.

If you insist on having width and height controls instead of just having a sane setting, then you _could_ just have width and height be variables that are used in the scaling logic. Then you will need to expose a scaling parameter per width and height too, as well as vertical and horizontal distance, and font size for title and label as well, else I think it will be quite useless.

I really think it's much better to output something that always works and letting the user scale a svg with correct proportions instead.

> Second, the different open pull requests that you made are focused on changing the same visualization structure so it would make sense to combine them and apply the changes to every single visualization plot. Otherwise, things might get a bit muddled.
I didn't think I would make more than one patch initially, oops. Agreed!

> Third, these changes will actually break the visualization code within BERTopic as the plotting functions are directly referenced within the main _bertopic.py code.
Oops, wasn't aware of that. I can fix it if you like.

> Having said that, I am currently working on v0.9.4 that will include fixes and some additional features. In this new version, I can look into making the automatic scaling available across all plotting functions whilst keeping the sizing parameters and perhaps let them inherit from an abstract plotting class to make sure all things are equal in basis across all plotting functions.
I'm not sure you can use the same scaling for all plotting functions, but maybe it works. I haven't read the ones that didn't break when I tried using them :)

MaartenGr:
> If you insist on having width and height controls instead of just having a sane setting

I am open to all and any feedback regarding BERTopic but be careful when phrasing it this way. To me, it seems that you are implying that what I propose is insane. Regardless of whether you agree or disagree with my approach, I would ask you to refrain from saying such things. 

I understand where you are coming from but width and height are most likely to remain. I do not make changes lightly to BERTopic and simply want to make sure that the changes I make fit with all users. Having said that, I will spend some time looking into this in either the next release or the one after that. For now, I believe I have some thinking to do.

simonfelding:
> > If you insist on having width and height controls instead of just having a sane setting
> 
> I am open to all and any feedback regarding BERTopic but be careful when phrasing it this way. To me, it seems that you are implying that what I propose is insane. Regardless of whether you agree or disagree with my approach, I would ask you to refrain from saying such things.
> 
> I understand where you are coming from but width and height are most likely to remain. I do not make changes lightly to BERTopic and simply want to make sure that the changes I make fit with all users. Having said that, I will spend some time looking into this in either the next release or the one after that. For now, I believe I have some thinking to do.

Hey, I'm really sorry you understand it that way, it was not my intention at all and I'll be more careful in the future. I don't think there's anything insane about your approach of course. "A sane default" was meant as "a default setting that works for the general purpose", not as implying that your idea is insane. Again I'm very sorry you took it like that, I think it's not an unusual expression and I meant no harm. I appreciate your approach to BERTopic very much and was just trying to suggest a way to achieve what you want.

I was just trying to say that I believe the default setting should work for more than 9 topics, and that adjusting width and height should behave like it does when you expand the size of a text editor: More words per line, instead of the words being stretched.

MaartenGr:
Glad to hear that you meant no harm. I was not familiar with the expression and my apologies for any misconception that resulted from that!

With respect to the sizing issue, I am more inclined to opt for the conservative method as it has more direct control over the sizing that users typically like to do directly from BERTopic and not so much from the resulting figure. There are a few other changes on the roadmap with respect to the figures so I will make sure to add this to them. For now, I believe it would be best to close this pull request either in favor of a new pull request or the changes that come with the upcoming version as we are starting to get quite far from the intention of the original pull request.

simonfelding:
Glad there's no conflic! Yes I can submit a new pull request with this format. But please reconsider if my initial suggestion isn't better, even though it slightly changes the api! I think my second solution is a bit hacky, but it should work fine. I just really think it makes more sense to specify how many columns you want instead of being forced to have 3 of variable size.

I would prefer to just implement this pull request with a scaled svg output - I truly believe it's a better solution. It would look like this:
`topic_model.visualize_barchart(top_n_topics = 16, n_words=10, columns=5, scale=2)`
So you don't have to specify the exact dimensions of the barchart, but instead just specify how large you want it without messing with dimensions. Fits my use case much better at least.

simonfelding:
I'm happy there's no conflict! Yes I can submit a new pull request with this format. While I think my second solution is a bit hacky, but it should work fine. But please reconsider if my initial suggestion isn't better, even though it slightly changes the api!  I just really think it makes more sense to specify how many columns you want instead of being forced to have 3 of variable size.

I would prefer to just implement this pull request with a scaled svg output - I truly believe it's a better solution. It would look like this:
`topic_model.visualize_barchart(top_n_topics = 16, n_words=10, columns=5, scale=2)`
So you don't have to specify the exact dimensions of the barchart, but instead just specify how large you want it without messing with dimensions. Fits my use case much better at least.

MaartenGr:
In that case, I would like to put this on hold for now. I understand your hesitance when it comes to the second solution but I have the same when it comes to your original solution. In the coming weeks, I hope to have some time testing out a few things and see what would fit with the api as it is right now. So a new pull request in the format of the second solution will also not be necessary as I am not sure of the approach that fits the package best.

simonfelding:
Alright, we'll have to wait and see and I'll run my own version until then :)

PR #343, opened 2021-11-24
Barplot x-axes are now the same length by default

simonfelding:
Before, x-axes were shared per column which was misleading. Now all x-axes are the same by default, or a range from 0-1 (sort of useful for finding extremely overfit topics).

Also added the option to change the title.


MaartenGr:
Closing this as discussed in https://github.com/MaartenGr/BERTopic/pull/342

PR #344, opened 2021-11-24
Automatically resize hierarchy plot to size.

simonfelding:
Now supports more topics. Like the other PRs I made. Doesn't need width and height input anymore, but left it there for tweaking purposes. I don't think it's necessary.


MaartenGr:
Closing this as discussed in https://github.com/MaartenGr/BERTopic/pull/342

PR #346, opened 2021-11-25
Tqdm autodetects if in a notebook environment or not

simonfelding:
It looks nicer this way and it chooses regular tqdm if it's not in a notebook.
[It's because my google colab notebook sometimes looks like this without it.](https://imgur.com/a/bSfpy4M)

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -13,7 +13,7 @@ import joblib
 import inspect
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


MaartenGr:
Thank you for the pr. Using auto_notebook may result in warnings that I want to prevent as much as possible. For now, the import will not be changed. However, I'll test it out in the coming weeks with tqdm.auto instead as the import can be found in several places.

Issue #349, opened 2021-11-30
`topics_per_class` raises an `IndexError`


HAMZA310:
### Issue

`topics_per_class` raises an `IndexError` **if `topics: List[int]` does not contain any outlier topic i.e. `-1`.**

I came across this issue while writing test cases for an NLP project I was working on based off of BERTopic.

Once discovered, the bug was easy to resolve. I'll create a PR with the quick fix.

### To Reproduce the Error:
I used a toy dataset, fitted BERTopic `topic_model`, and then called `topic_model.topics_per_class`.
```
# Toy dataset I used:
n_copies = 5
docs = [
    ['GLOBAL STANDARD FOOD']*n_copies,
    ['BRCGS 5 Global Stand']*n_copies,
    ['Coles Supplier Requi']*n_copies,
    ['Export Control (Milk']*n_copies,
    ['2 Compilation date: ']*n_copies,
    ['FSSC 22000 SCHEME VE']*n_copies,
    ['2345 Crystal Drive S']*n_copies,
    ['Food Safety Code: St']*n_copies,
    ['Raw Material Managem']*n_copies,
    ]
```

As topics are not deterministic, outlier topics will sometimes be present/absent in an execution.

MaartenGr:
Good catch, thanks! You actually stumbled upon an issue I am not sure how high of a priority it should get. Since no outlier class is present, and I assume throughout BERTopic's code that it is, there are likely to be more places where issues may pop up. This would mean that there might be a bunch more cases where changes need to be made making this into quite a large fix for a use case (very small data) it not really is intended for. 

Having said that, I am close to finishing v0.9.4 in the coming weeks, so I'll take a closer look at the PR you made and see what the implications are for supporting very small datasets.

HAMZA310:
I agree.

I've been using BERTopic in my project for a while now, and I also assumed the presence of outlier topics until I wrote an integration test case. I've handled it awkwardly in my project `testing` code for now. Will look forward to v0.9.4 for a more robust fix.

Have fun working on next release!

MaartenGr:
I am going to close this issue for now as I believe that the fix would allow for a use case it is not really intended for. However! If there are more users running into this issue and have an idea for a fix across the entire package, please let me know!

HAMZA310:
> have an idea for a fix across the entire package


Yes, I can create a PR for that if needed? Yet, can't promise *when* it'll be ready.

MaartenGr:
[...]

PR #350, opened 2021-11-30
Fix `topics_per_class raises an IndexError if no outlier topics`

HAMZA310:
ref #349


MaartenGr:
Sorry for the late response! I am going to close this for now as discussed in #349. However, I will put this on the roadmap as it likely affects quite a few functions and features.

PR #390, opened 2022-01-01
to include optimal ordering dendogram, as https://github.com/rafaelva…

rafaelvalero:
…lero/different_notebooks/blob/master/bertopics_hierarchy.py

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -1103,7 +1103,8 @@ class BERTopic:
                             topics: List[int] = None,
                             top_n_topics: int = None,
                             width: int = 1000,
-                            height: int = 600) -> go.Figure:
+                            height: int = 600,
+                            optimal_ordering: bool = False) -> go.Figure:
         """ Visualize a hierarchical structure of the topics
 
         A ward linkage function is used to perform the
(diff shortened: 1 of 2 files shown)


MaartenGr (changes requested):
Could you also add the `optimal_ordering` parameter, including docstring in `_bertopic.py`? You can find it in the permalink below: 

https://github.com/MaartenGr/BERTopic/blob/cd98fc8d22ab1eba593c518278ce479d2879c372/bertopic/_bertopic.py#L1101-L1106

The functions in the plotting backend are directly used in `_bertopic.py`, so any changes there will also need to be made in the model itself as it currently does not inherit the entire function.

rafaelvalero:
Thanks for the feed-back, I did not realise it was there too. I think it is done now.

bertopic/_bertopic.py
```diff
@@ -1117,7 +1118,11 @@ def visualize_hierarchy(self,
             top_n_topics: Only select the top n most frequent topics
             width: The width of the figure. Only works if orientation is set to 'left'
             height: The height of the figure. Only works if orientation is set to 'bottom'
-
+            optimal_ordering: bool, optional. If True, the linkage matrix will be reordered so that the distance
```
MaartenGr, line 1121: Could you remove the `bool, optional.` here? Thus far, BERTopic only mentions the typing when declaring the parameters and not in the documentation itself.

bertopic/plotting/_hierarchy.py
```diff
@@ -27,7 +28,11 @@ def visualize_hierarchy(topic_model,
         top_n_topics: Only select the top n most frequent topics
         width: The width of the figure. Only works if orientation is set to 'left'
         height: The height of the figure. Only works if orientation is set to 'bottom'
-
+        optimal_ordering: bool, optional. If True, the linkage matrix will be reordered so that the distance
```
MaartenGr, line 31: Same here. Other than that it looks good. It might take some time for this to be merged as there are currently some issues with the pipeline. I will, however, starting running the checks.

MaartenGr:
Apologies for the late reply! Finally had some time to go over the pull requests that were open and this looks good to me, so it can be merged.

PR #421, opened 2022-01-25
Support custom labels for plots

simonfelding:
I think this is a very useful feature. I added functions to bertopic.py that exports an xlsx sheet with useful data for labeling, but it's not important. Thought it would be nice for usability though.

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -881,6 +881,7 @@ class BERTopic:
     def visualize_topics(self,
                          topics: List[int] = None,
                          top_n_topics: int = None,
+                         labels: List[str] = None,
                          width: int = 650,
                          height: int = 650) -> go.Figure:
         """ Visualize topics, their sizes, and their corresponding words
@@ -891,6 +892,7 @@ class BERTopic:
         Arguments:
             topics: A selection of topics to visualize
             top_n_topics: Only select the top n most frequent topics
+            labels: List of custom labels to use.
             width: The width of the figure.
             height: The height of the figure.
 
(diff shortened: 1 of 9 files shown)


simonfelding:
Screenshot 2022-01-25 at 04 58 40

It looks like this now.

MaartenGr:
Thank you for the PR! The changes you made are close to a feature that is on the roadmap, namely creating custom topic representations. As you mentioned, based on the topic representations, users might want to give them a custom name. However, I do think that it would be cleaner to create those custom representations/names within BERTopic instead of the need to pass lists back and forth. That way, they are internally represented and can be used across a variety of functions and more than just the visualizations.

simonfelding:
You could easily modify it to be an internal state instead, but I feel like it's something that's nice to keep seperate from the model itself, since it's something that you might have to change if you realize you did something wrong, and turn on and off easily. For example in my analysis, I had some topics I didn't know how to clearly define, but when I put them in the hierarchial plot it became clear what it was.

I don't really see how you would create the labels in BERTopic without making a complicated analysis and labeling UI, which is something that excel already does really well in my opinion. But it's a good idea to be able to pass the labels to functions like get_topics() and so on! But I think the most powerful analysis features are the visualisations.

But okay, maybe the internal labels could just be an object initialized as an empty list, set like ´topic_model.labels = some_list´ and automatically passed to everything. Then the user gets to figure out how to generate a list of labels without using my handy excel export function. For me it makes sense to keep it seperate though, but I guess I'm just fond of stateless programming. The biggest problem for me with the stateful version would be how to revert the state back to unlabeled without storing the old labels and the custom labels at the same time and having some switching mechanism to turn it on and off. On the other hand, BERTopic could just regenerate the labels if it's an empty list.. I don't like this approach much, maybe you have a better idea.

MaartenGr:
In my experience, users typically want these things to be saved within BERTopic, and this is something I definitely would prefer as well. Ideally, the mapping can be generated by defining a dictionary that maps one topic label to another without the need for excell which might not be available to all users. 

For now, I'll put this on the roadmap to give me some time to think about how this fits best within the current design thinking and choices surrounding BERTopic.

PR #422, opened 2022-01-25
Dendrogram color threshold

simonfelding:
If this parameter isn't exposed, the dendrogram will often generate plots with inexplicable colors. I personally prefer setting it to 2, but 1 was set before so I'm assuming it's a good default setting.

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -1102,6 +1102,7 @@ class BERTopic:
                             orientation: str = "left",
                             topics: List[int] = None,
                             top_n_topics: int = None,
+                            color_threshold: float = 1,
                             width: int = 1000,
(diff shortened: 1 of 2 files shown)


simonfelding:
Screenshot 2022-01-25 at 04 58 40

It looks like this now (including my custom labels patch).

MaartenGr:
Thank you for the PR. Exposing the `color_threshold` variable does indeed seem like it could generate a nicer representation of the clusters. I would need some time to test this out and a few of the other visualization changes you made before merging. Also, there are currently some ABI issues with HDBSCAN and Numpy which prevents me from running a clean workflow so it might take some time before it gets merged.

MaartenGr:
Seeing as the hierarchical visualizations have changed quite a bit with the latest release, I will be closing this PR.

Issue #383, opened 2021-12-26
Huge model with save() and large dataset


yotammarton:
Hello,
Saving a BERTopic model trained with 3.5M texts, the resulting file is some ±25GB big.
I have a memory (RAM) concerns when inferencing with the model (*).

I am trying to lower the model size by (possibly) loading it, manipulating its inner `self` variables (maybe None some of them?) and save it again, hoping for a smaller model.

1. I actually don't use in my application any of the cTFIDF, is it possible to delete `self.c_tf_idf` for example?
2. What are all the big variables and matrices I am able to delete (and how to do so correctly? `None` or `del self.x`) in order to reduce the model size? Referring to HDBSCAN and UMAP as well if you know. How complicated is that?
3. Regarding Q1. a bit irrelevant question to the subject, but anyway.. How BERTopic is able to classify correctly (to the expected cluster, name `c`) a sentence that is only Out-of-Vocabulary words for my `embedding_model`? e.g. `all-MiniLM-L6-v2` model and sentence "mothersday" is classified to the expected cluster `c`, even though `mothersday` is not part of the vocabulary (but is a top-10 cTFIDF word in the `c`)

(*) I am creating an inference server that creates replicas of the model to reduce the inference time, I want as many replicas as possible. For 2 replicas I need 2 * 25GB RAM machine.

Thank you :)

MaartenGr:
In general, I would not recommend removing some of the `self` as they all have a specific purpose and are used throughout the model. Do this only if you are absolutely sure yourself that it is not interfering with any of the functionality of the model. 

Identifying why the model is large in the first place is key and is something you should do before looking into deleting certain variables. There are a number of sub-models that can potentially be quite large:
* Embedding model
* HDBSCAN
* UMAP
* c-TF-IDF

## Example
To give you an idea of the effect of certain models:

```python
from bertopic import BERTopic
from sklearn.datasets import fetch_20newsgroups

docs = fetch_20newsgroups(subset='all',  remove=('headers', 'footers', 'quotes'))['data']

topic_model = BERTopic(verbose=True)
topics = topic_model.fit_transform(docs)
topic_model.save("full_model")
```

Then, you can remove certain models from BERTopic to check its effect on the model size:

```python
umap_model = topic_model.umap_model
topic_model.umap_model = None
topic_model.save("without_umap")
```

[...]

PR #440, opened 2022-02-13
don't save the fitted vectorizer model in the model!

simonfelding:
A fix to https://github.com/MaartenGr/BERTopic/issues/383

Storing a fitted `vectorizer_model` makes the topic model extremely memory hungry. This way, `self._c_tf_idf` is slower, but it's anyway only ever sped up in the second run in the case of doing `topics_over_time` or `topics_per_class` *after* already fitting the model once.

CountVectorizer is pretty fast already (takes about 2 minutes on my 48.000 document dataset), so I don't think it's worth storing the entire fitted model just to `.get_feature_names()` a little faster in special use cases compared to the cost in memory. It's in most use cases only fitted once, and the fitting stage isn't time-sensitive either. Plus, the `.transform(documents)` that had to be done anyway still fits first anyway, so the speedup isn't so big in the first place.

Also updated `.get_feature_names()` to `.get_feature_names_out()`, as the former is deprecated and will stop working with sklearn 1.2, which will be out soon.

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -455,6 +455,8 @@ class BERTopic:
         check_documents_type(docs)
         documents = pd.DataFrame({"Document": docs, "Topic": topics, "Timestamps": timestamps})
         global_c_tf_idf = normalize(self.c_tf_idf, axis=1, norm='l1', copy=False)
+        vectorizer_model = clone(self.vectorizer_model).fit(documents)
+        c_tf_idf_model = 0 # gets fitted later
 
         all_topics = sorted(list(documents.Topic.unique()))
         all_topics_indices = {topic: index for index, topic in enumerate(all_topics)}
@@ -485,7 +487,7 @@ class BERTopic:
             selection = documents.loc[documents.Timestamps == timestamp, :]
             documents_per_topic = selection.groupby(['Topic'], as_index=False).agg({'Document': ' '.join,
                                                                                     "Timestamps": "count"})
-            c_tf_idf, words = self._c_tf_idf(documents_per_topic)
+            c_tf_idf, words, c_tf_idf_model  = self._c_tf_idf(documents_per_topic, vectorizer_ = vectorizer_model, c_tf_idf_ = c_tf_idf_model)
 
             if global_tuning or evolution_tuning:
                 c_tf_idf = normalize(c_tf_idf, axis=1, norm='l1', copy=False)
@@ -569,7 +571,9 @@ class BERTopic:
         """
         documents = pd.DataFrame({"Document": docs, "Topic": topics, "Class": classes})
         global_c_tf_idf = normalize(self.c_tf_idf, axis=1, norm='l1', copy=False)
-
+        vectorizer_model = clone(self.vectorizer_model).fit(documents)
+        c_tf_idf_model = 0 # gets fitted later
+        
         # For each unique timestamp, create topic representations
         topics_per_class = []
         for index, class_ in tqdm(enumerate(set(classes)), disable=not self.verbose):
@@ -578,7 +582,7 @@ class BERTopic:
             selection = documents.loc[documents.Class == class_, :]
             documents_per_topic = selection.groupby(['Topic'], as_index=False).agg({'Document': ' '.join,
                                                                                     "Class": "count"})
-            c_tf_idf, words = self._c_tf_idf(documents_per_topic)
+            c_tf_idf, words, c_tf_idf_model = self._c_tf_idf(documents_per_topic, vectorizer_ = vectorizer_, c_tf_idf_ = c_tf_idf_model)
 
             # Fine-tune the timestamp c-TF-IDF representation based on the global c-TF-IDF representation
             # by simply taking the average of the two
@@ -1562,24 +1566,32 @@ class BERTopic:
 
             self.topic_embeddings = topic_embeddings
 
-    def _c_tf_idf(self, documents_per_topic: pd.DataFrame) -> Tuple[csr_matrix, List[str]]:
+    def _c_tf_idf(self, documents_per_topic: pd.DataFrame, **models) -> Tuple[csr_matrix, List[str]]:
         """ Calculate a class-based TF-IDF where m is the number of total documents.
 
         Arguments:
             documents_per_topic: The joined documents per topic such that each topic has a single
                                  string made out of multiple documents
             m: The total number of documents (unjoined)
+            return_fitted: Also return fitted vectorizer and c_tf_idf 
+            vectorizer_: A pre-fitted CountVectorizer instance. Optional.
+            c_tf_idf_: A pre-fitted c_tf_idf instance. Optional.
 
         Returns:
             tf_idf: The resulting matrix giving a value (importance score) for each word per topic
             words: The names of the words to which values were given
+            c_tf_idf_: The fitted c_tf_idf instance (If **models is not empty)
         """
         documents = self._preprocess_text(documents_per_topic.Document.values)
 
-        fitted_vectorizer = clone(self.vectorizer_model).fit(documents)
+        if "vectorizer_" in models:
+            vectorizer_ = models["vectorizer_"]
+        else:
+            vectorizer_ = clone(self.vectorizer_model)
+            vectorizer_.fit(documents)
 
-        words = fitted_vectorizer.get_feature_names_out()
-        X = fitted_vectorizer.transform(documents)
+        words = vectorizer_.get_feature_names()
+        X = vectorizer_.transform(documents)
 
         if self.seed_topic_list:
             seed_topic_list = [seed for seeds in self.seed_topic_list for seed in seeds]
@@ -1587,12 +1599,18 @@ class BERTopic:
         else:
             multiplier = None
 
-        self.transformer = ClassTFIDF().fit(X, multiplier=multiplier)
+        if "c_tf_idf_" in models and models['c_tf_idf_'] != 0:
+            c_tf_idf_ = models["c_tf_idf_"]
+        else:
+            c_tf_idf_ = ClassTFIDF().fit(X, multiplier=multiplier)
 
+        c_tf_idf = c_tf_idf_.transform(X)
 
         self.topic_sim_matrix = cosine_similarity(c_tf_idf)
-
-        return c_tf_idf, words
+        if models not {}:
+            return c_tf_idf, words, c_tf_idf_
+        else:
+            return c_tf_idf, words
 
     def _update_topic_size(self, documents: pd.DataFrame):
         """ Calculate the topic sizes
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -21,6 +21,7 @@ from typing import List, Tuple, Union, Mapping, Any
 import hdbscan
 from umap import UMAP
 from sklearn.feature_extraction.text import CountVectorizer
+from sklearn.base import clone
 from sklearn.metrics.pairwise import cosine_similarity
 from sklearn.preprocessing import normalize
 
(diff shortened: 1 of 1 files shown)


simonfelding:
after some testing and reading, it turned out that you need to use `sklearn.base.clone` to copy the `vectorizer_model` in order to avoid updating the base function saved in the model. Now it works as intended and doesn't store the huge fitted `vectorizer_model` permanently.

simonfelding:
now the fitted vectorizer isn't stored anymore, there's also no need to store the fitted ctfidf either (as `self.transformer`) in my opinion. It's also quite large, but maybe you want to keep it for some reason.

bertopic/_bertopic.py
```diff
@@ -1561,35 +1562,32 @@ def _create_topic_vectors(self):
...
             seed_topic_list = [seed for seeds in self.seed_topic_list for seed in seeds]
             multiplier = np.array([1.2 if word in seed_topic_list else 1 for word in words])
         else:
             multiplier = None
 
-        if fit:
-            self.transformer = ClassTFIDF().fit(X, multiplier=multiplier)
+        self.transformer = ClassTFIDF().fit(X, multiplier=multiplier)
 
         c_tf_idf = self.transformer.transform(X)
```
simonfelding, line 1592:
```suggestion
        c_tf_idf = ClassTFIDF().fit(X, multiplier=multiplier).transform(X)
```

simonfelding, line 1592:
```suggestion

```

MaartenGr:
Although I certainly appreciate the effort, I would strongly recommend discussing a solution with the maintainer before working on an issue. You risk doing work that the maintainer might not necessarily agree with. In this particular case, it is important that the fitted vectorizer model stays in place as that specific fitted instance is used later on in `topics_over_time`  and `topics_per_class`. The reason for that is that re-fitting them in `topics_over_time` generates a completely different bag-of-words representation than would be the case if I were to use the fitted vectorizer model.

simonfelding:
I changed it now so `topics_per_class` and `topics_over_time` have their own cloned instances of countvectorizer that can be used multiple times. It still unecessarily avoids storing the countvectorizer permanently in ram, but without breaking `topics_per_class` and `topics_over_time`.

MaartenGr:
> Or do both functions need to have the exact same vocabulary (the data generated when the counvectorizer is fitted)? I see them as entirely seperated from each other, so I think it's enough that each of them just generate and store their own vocabulary when they are run until they are finished.

Yes, we fit the CountVectorizer once to get a bag-of-words representation that can then be used in `topics_per_class` and `topics_over_time`. These need to have the exact same vocabulary otherwise it becomes difficult to make comparisons between the main topic representation and the topic representations generated by those functions.

simonfelding:
Great, then it works as intended! I checked if they are the same, and they are.
Screenshot 2022-02-14 at 16 46 23

MaartenGr:
Perhaps I have been unclear before but I want to keep a fitted CountVectorizer in BERTopic. Not just for the two previously mentioned functions but because it is in general very handy to save that model for uses surrounding topic modeling, such as the calculation of coherence and covariance significance.

simonfelding:
ok. as an alternative, it can be stored compressed when not in use.

I checked what compression algorithms are best suited:
```
initial model size: 891.597MB type: CountVectorizer
zstandard compression:
	size: 82.801MB
	time to compress: 4.64s
	datatype: bytes
	time to decompress 3.91s. roundtrip: 8.55s
	decompressed datatype: CountVectorizer
--------

zlib compression:
	size: 75.183MB
	time to compress: 19.52s
	datatype: bytes
	time to decompress 4.75s. roundtrip: 24.27s
	decompressed datatype: CountVectorizer
--------

lz4.frame compression:
	size: 118.403MB
	time to compress: 3.75s
	datatype: bytes
	time to decompress 3.49s. roundtrip: 7.24s
	decompressed datatype: CountVectorizer
--------

bz2 compression:
	size: 62.882MB
	time to compress: 23.85s
	datatype: bytes
	time to decompress 15.19s. roundtrip: 39.04s
	decompressed datatype: CountVectorizer
--------
```

I think zstandard is the best compromise between speed and size, but I think you will prefer zlib. zlib and bz2 are standard libraries, so they won't require a change in requirements for bertopic. Of these, zlib is best because it decompresses almost as fast as zstd, and gets better compression too.

I'll submit the code to make it work.

MaartenGr:
Thank you for diving into that solution. But again, let's take a step back first and discuss the issue before proposing and working on solutions. You risk focusing on things and doing work that might not fit with the design and preferences that I have in mind. In this particular case, I actually have not been able to reproduce your issue, so creating a PR with a specific solution in mind may have been a bit hasty. 

When I run the following for a dataset of 46694 documents, the resulting CountVectorizer size is actually the lowest:

```python
import re
import pandas as pd
from bertopic import BERTopic
from pympler.asizeof import asizeof

# Prepare data
trump = pd.read_csv('https://drive.google.com/uc?export=download&id=1xRKHaP-QwACMydlDnyFPEaFdtskJuBa6')
trump = trump.loc[(trump.isRetweet == "f") & (trump.text != ""), :]
data = trump.text.to_list()

# Train model
topic_model = BERTopic()
topics, probs = topic_model.fit_transform(data)

# Calculate size
def bytesto(bytes, to, bsize=1024): 
    a = {'k' : 1, 'm': 2, 'g' : 3, 't' : 4, 'p' : 5, 'e' : 6 }
    r = float(bytes)
    return bytes / (bsize ** a[to])
print(f"vectorizer: {bytesto(asizeof(topic_model.vectorizer_model), 'm')}'M'")
print(f"hdbscan: {bytesto(asizeof(topic_model.hdbscan_model), 'm')}'M'")
print(f"umap: {bytesto(asizeof(topic_model.umap_model._raw_data), 'm')}'M'")
print(f"c-tf-idf: {bytesto(asizeof(topic_model.c_tf_idf), 'm')}'M'")
```

With the following results:

```
vectorizer: 7.705741882324219'M'
hdbscan: 10.482872009277344'M'
umap: 68.3995361328125'M'
c-tf-idf: 2.6476898193359375'M'
```

In this case, I do not see the CountVectorizer being that large. It might grow larger if you set the n-gram-range bigger than 5 but in general, that should not be necessary. 

> I'll submit the code to make it work, unless you don't like this approach either.

My apologies but that is very much a possibility if you start working on an approach without discussing it first. You have a tendency to start working on a solution before actually agreeing on that solution. I want to make clear that I do not enjoy saying no but that I also will not implement any solutions I do not agree with, even if I have to say no more than I would like.

simonfelding:
Thanks for your response. Yes I just politely try to make it easy for you, that's the procedure I'm used to: Contribute a working fix, change it from discussion and using the review function until it's acceptable.

The reason you cannot replicate with that dataset is that it doesn't have a lot of different words because you are using Trump's texts.

[Try with this instead](https://drive.google.com/file/d/1gfq3rtX_mPv_2hOzefZBHone0V4KqOad/view?usp=sharing).
```
import pandas as pd
from bertopic import BERTopic
from pympler.asizeof import asizeof

# import
data = pd.read_feather('bpd_df.feather').titlebody.tolist()

# Train model
topic_model = BERTopic()
topics, probs = topic_model.fit_transform(data)

# Calculate size
def bytesto(bytes, to, bsize=1024): 
    a = {'k' : 1, 'm': 2, 'g' : 3, 't' : 4, 'p' : 5, 'e' : 6 }
    r = float(bytes)
    return bytes / (bsize ** a[to])
print(f"vectorizer: {bytesto(asizeof(topic_model.vectorizer_model), 'm')}'M'")
print(f"hdbscan: {bytesto(asizeof(topic_model.hdbscan_model), 'm')}'M'")
print(f"umap: {bytesto(asizeof(topic_model.umap_model._raw_data), 'm')}'M'")
print(f"c-tf-idf: {bytesto(asizeof(topic_model.c_tf_idf), 'm')}'M'")
```

MaartenGr:
> Thanks for your response. Yes I just politely try to make it easy for you, that's the procedure I'm used to: Contribute a working fix, change it from discussion and using the review function until it's acceptable.

That is absolutely appreciated! Just make sure to thoroughly discuss the issue before starting on a fix. Your latest commit stresses the point that I am trying to make. We did not agree yet on compression as a way to solve large CountVectorizers yet you continue working on it as if disregarding the points that I am trying to make. 

> The reason you cannot replicate with that dataset is that it doesn't have a lot of different words, and the documents are very small, because you are using Trump's tweets.

In that case, and taking a step back looking at the issue, it seems to stem mostly from `.stop_words_` according to sklearn's [documentation](https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html):

> The `stop_words_` attribute can get large and increase the model size when pickling. This attribute is provided only for introspection and can be safely removed using delattr or set to None before pickling.

In other words, it should be resolved by letting the user set a larger `min_df` and by removing the `stop_words_` before saving the vectorizer. Seeing as it will be a one-liner, I am going to implement that in the next release.

simonfelding:
> > The `stop_words_` attribute can get large and increase the model size when pickling. This attribute is provided only for introspection and can be safely removed using delattr or set to None before pickling.
> 
> In other words, it should be resolved by letting the user set a larger `min_df` and by removing the `stop_words_` before saving the vectorizer. Seeing as it will be a one-liner, I am going to implement that in the next release.

No, it's not actually the stopwords - it's the enormous vocabulary in this case. I checked it :) Limiting stopwords is not the solution for this problem, sadly. The vocabulary can get large too:

https://scikit-learn.org/stable/modules/feature_extraction.html#vectorizing-a-large-text-corpus-with-the-hashing-trick
> The above vectorization scheme is simple but the fact that it holds an in- memory mapping from the string tokens to the integer feature indices (the vocabulary_ attribute) causes several problems when dealing with large datasets:
> the larger the corpus, the larger the vocabulary will grow and hence the memory use too,

MaartenGr:
> [Look at this commit: It automatically compresses all attributes set in the bertopic class.](https://github.com/MaartenGr/BERTopic/pull/440/commits/76a52a28f8f449163a5b80bb9acd4c7d97f2a03c)

That's exactly what I mean, I do not agree on compression when there are other solutions out there that better fit within the current state of BERTopic, including its design and execution. That is also the commit that I was referring to in my previous response. 

> No, it's not actually the stopwords - it's the enormous vocabulary in this case. I checked it :) Limiting stopwords is not the solution for this problem, sadly. The vocabulary can get large too:

That's the thing, by setting a larger `min_df`, for example `2`, it decreases the vocabulary by ignoring all words that have a frequency of less than 2. However, it puts all words that have a frequency of less than 1 in the `stop_words_` variable. As a result, it only takes a single line to remove those stop words as users already have to take into account the generation of large c-TF-IDF matrices by working with `min_df`. 

However, I feel like we are talking in circles and it does not seem we will end on an agreement. As I mentioned before, compression needs some testing but is still on the roadmap when it concerns saving BERTopic but will not be implemented as of right now for the CountVectorizer.

simonfelding:
ok - but the problem is the vocabulary, not the stopwords. i'll continue running my patched version. I really don't understand why you think compression needs to be tested more than this.

Maybe you could leave this pull request open so people can go here in the future and get my patch.

simonfelding:
Just a proof of what I'm saying:
`topic_model.vectorizer_model.stop_words_`
`set()`
It's an empty set, so definitely not the problem.

MaartenGr:
> Maybe you could leave this pull request open so people can go here in the future and get my patch.

I understand that it is a solution that works very well for you and I am happy to hear that! However, I am not going to leave a pull request open that I do not agree with. 

> I also think it's a bit rude to implement it yourself later like you did with some of my fixes for the visualization code, because it means that I don't get any credit for my work.

You are absolutely correct in saying that it would be rude to implement what you did in this PR myself later. Perhaps what I said was a bit misleading and I apologize for that but I was referring to the pull request that you made that is currently still open. That is the one that is still on the roadmap and might be implemented later on through a merge of your pull request. The solution that I foresee, namely working with the `.stop_words_` variable is something I will likely be implementing myself.

PR #448, opened 2022-02-16
fix RuntimeError when used as sklearn estimator

simonfelding:
Simple fix for a simple bug:
```RuntimeError: scikit-learn estimators should always specify their parameters in the signature of their __init__ (no varargs). <class 'bertopic._ctfidf.ClassTFIDF'> with constructor (self, *args, **kwargs) doesn't  follow this convention.```

It doesn't take any arguments in `__init__`, so there's no need for taking varargs. There are also no arguments used in the _bertopic.py code.

### DIFF
--- a/bertopic/_ctfidf.py
+++ b/bertopic/_ctfidf.py
@@ -18,8 +18,8 @@ class ClassTFIDF(TfidfTransformer):
     Next, the total, unjoined, number of documents across all classes **m** is divided by the total
     sum of word **i** across all classes.
     """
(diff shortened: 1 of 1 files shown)


simonfelding:
The init doesn't actually do anything so there's not really any need for it. It doesn't store any state information, so the entire `def __init__():` could just as well be removed. it currently just inherits it's own init, which is empty.

MaartenGr:
Thank you for the PR. LGTM!

PR #475, opened 2022-03-15
Fix minor typo

dwhdai:
There was a minor typo in the Visualization docs :)


MaartenGr:
That's a horrible typo 😅 Thanks for fixing it! I'll regenerate the documentation most likely when the new version is released, so it will take some time before the typo is fixed in the live documentation.

PR #566, opened 2022-06-16
Add sorting to Topic order

proselotis:
Adding sorting in the visualize_topics_over_time function before plotting the topics will ensure the legend looks clean and orderly by having the most popular topics at the top. This will also save the user time from having to look through the legend by ordering in a more intuitive way rather than having the user scroll through the legend when looking to select or deselect a topic.

Screenshots for example provided:

Previous:
previous_visualize_topics_over_time

Now:
sorted_visalize_topics_over_time

### DIFF
--- a/bertopic/plotting/_topics_over_time.py
+++ b/bertopic/plotting/_topics_over_time.py
@@ -58,7 +58,7 @@ def visualize_topics_over_time(topic_model,
     topic_names = {key: value[:40] + "..." if len(value) > 40 else value
                    for key, value in topic_model.topic_names.items()}
     topics_over_time["Name"] = topics_over_time.Topic.map(topic_names)
-    data = topics_over_time.loc[topics_over_time.Topic.isin(selected_topics), :].sort_values(["Topic","Timestamp"])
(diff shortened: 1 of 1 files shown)


bertopic/plotting/_topics_over_time.py
```diff
@@ -58,8 +58,8 @@ def visualize_topics_over_time(topic_model,
     topic_names = {key: value[:40] + "..." if len(value) > 40 else value
                    for key, value in topic_model.topic_names.items()}
     topics_over_time["Name"] = topics_over_time.Topic.map(topic_names)
-    data = topics_over_time.loc[topics_over_time.Topic.isin(selected_topics), :]
-
+    data = topics_over_time.loc[topics_over_time.Topic.isin(selected_topics), :].sort_values(["Topic","Timestamp"])
```
MaartenGr, line 61: Very very minor change, could you put an additional space between after the `,` in `"Topic","Timestamp"`? 
Other than that, it looks good to me, thanks for the fix. 

```suggestion
    data = topics_over_time.loc[topics_over_time.Topic.isin(selected_topics), :].sort_values(["Topic", "Timestamp"])
```

proselotis:
Suggestions understood and addded! Thanks for the quick response

MaartenGr:
Awesome, thanks. LGTM!

PR #617, opened 2022-07-15
Support custom options for all backend models.

### LINKED ISSUE #387: Setting batch_size parameter for encoding embeddings with Sentence transformers
I am currently using my laptop's GPU to encode embeddings and because I have to pass it externally to the BertTopic class since the default 32 batch size will cause a memory overflow. Don't know if there is a way we can pass batch size straight into the BertTopic class or the fit_transform functions somehow? 

```python
 model = SentenceTransformer("paraphrase-multilingual-mpnet-base-v2", device="cuda")
 embeddings = model.encode(docs, show_progress_bar=True, batch_size=16)
 topic_model = BERTopic(embedding_model=model, calculate_probabilities=True)
```

zhawe01:
### What does this PR do?
Fixes #387 
Support custom options, for example 'batch_size', for all backend models. Here is an example:

> ```
> model = SentenceTransformer(model_name)
> embeddings = model.encode(docs, batch_size=16)
> topic_model = BERTopic(embedding_model=model, calculate_probabilities=True)
> topic_model.fit_transform(docs)
> ```
> After supporting custom options, the above code snippet can be simplified to the following: 
> ```
> topic_model = BERTopic(embedding_model=model_name, calculate_probabilities=True)
> predictions, probs = topic_model.fit_transform(docs, batch_size=16)
> ```

### Who can review?
@MaartenGr

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -198,7 +198,8 @@ class BERTopic:
     def fit(self,
             documents: List[str],
             embeddings: np.ndarray = None,
-            y: Union[List[int], np.ndarray] = None):
+            y: Union[List[int], np.ndarray] = None,
+            **kwargs):
         """ Fit the models (Bert, UMAP, and, HDBSCAN) on a collection of documents and generate topics
 
         Arguments:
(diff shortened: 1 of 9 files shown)


MaartenGr:
Thank you for your work on this PR! Although this makes things easier in that use case, I want to prevent the usage of `**kwargs` arguments as much as possible throughout BERTopic unless it is absolutely necessary. Having `**kwargs` can make the API much more difficult to read and understand, since the `**kwargs` can be used at several different places within and across functions.  I believe that in these cases, it might reduce the transparency of parameters as it is not explicit where and how they are used.

zhawe01:
> 

Thank you for the quick response. I really understand your concern. Hopefully a more balanced solution can be found in the future.

PR #660, opened 2022-08-04
fix: BERTopic. get_topic_info null condition

### LINKED ISSUE #659: BERTopic.get_topic_info(topic=0) return the entire info dataframe for all topics
the following condition
```
if topic:
        info = info.loc[info.Topic == topic, :]
```
is too ambiguous and will evaluate to False if `topic == 0`

oxymor0n:
fixes #659


MaartenGr:
Good catch, thanks for the PR!

PR #734, opened 2022-09-21
fixed typo vehicle -> gpu

timpal0l:


MaartenGr:
It seems that the checks were not successful due to HDBSCAN. Hopefully, that package gets updated soon so that I can properly run the workflow.

Issue #768, opened 2022-10-07
Would you be open to a lightweight vectorizer/embedder? 


koaning:
The datasets that I have tend to be ~80K examples and just running the embeddings on a CPU takes ~40 minutes. 

I have, however, a trick up my sleeve. 

```python
from sklearn.pipeline import make_pipeline, make_union
from sklearn.decomposition import TruncatedSVD
from sklearn.feature_extraction.text import HashingVectorizer, TfidfTransformer

pipe = make_pipeline(
    make_union(
        HashingVectorizer(n_features=10_000),
        HashingVectorizer(n_features=9_000),
        HashingVectorizer(n_features=8_000)
    ),
    TfidfTransformer(),
    TruncatedSVD(100)
)
```

This pipeline combines a hashing trick, with a bloom hack, with a sparse pca trick and a tf-idf trick. One benefit is that this is orders of magnitude faster to embed; even when you include training. 

```python
import numpy as np
import perfplot
from sentence_transformers import SentenceTransformer

sentence_model = SentenceTransformer("all-MiniLM-L6-v2")
out = perfplot.show(
    setup=lambda n: docs[:n],  # or setup=np.random.rand
    kernels=[
        lambda a: pipe.fit_transform(a),
        lambda a: pipe.fit(a).transform(a),
        lambda a: sentence_model.encode(a)
    ],
    labels=["fit_transform", "fit_and_transform", "sentence_transformer"],
    n_range=[100, 200, 500, 1000, 2000, 5000, 10_000],
    xlabel="len(a)",
    equality_check=None
)
```

![image](https://user-images.githubusercontent.com/1019791/194563105-3847d6d9-3c0c-4584-bec7-2e5067aa9907.png)

It's just orders of magnitude faster. So maybe it'd be nice to have these embeddings around? 
 
## But what about the quality of the embeddings? 

Mileage can vary, sure, but I have some results [here](https://koaning.github.io/cluestar/) that suggest it's certainly not the worst idea either. When you compare the UMAP chart on top of tf/idf with the universal sentence encoder one then, sure ... the USE variant is intuitively better, but given the speedup, I might argue that the tf/idf approach is reasonable too. 

There's a fair bit of tuning involved, and I'm contemplating a library that implements bloom vectorizers properly for scikit-learn. But once that is done and once I've done some benchmarking, would this library be recipient to such an embedder?

MaartenGr:
[...]

PR #791, opened 2022-10-21
Sklearn Pipeline Embedder

koaning:
This PR fixes https://github.com/MaartenGr/BERTopic/issues/768.

### DIFF
--- a/bertopic/backend/_sklearn.py
+++ b/bertopic/backend/_sklearn.py
@@ -53,7 +53,7 @@ class SklearnEmbedder(BaseEmbedder):
 
         Arguments:
             documents: A list of documents or words to be embedded
-            verbose: Controls the verbosity of the process
+            verbose: No-op variable that's kept around to keep the API consistent. If you want to get feedback on training times, you should use the sklearn API.
 
         Returns:
             Document/words embeddings with shape (n, m) with `n` documents/words
--- a/bertopic/backend/_sklearn.py
+++ b/bertopic/backend/_sklearn.py
@@ -1,0 +1,68 @@
+from bertopic.backend import BaseEmbedder
+from sklearn.utils.validation import check_is_fitted, NotFittedError
+
+
+class SklearnEmbedder(BaseEmbedder):
+    """ Scikit-Learn based embedding model
+
+    This component allows the usage of scikit-learn pipelines for generating document and
+    word embeddings.
+
+    Arguments:
+        pipe: A scikit-learn pipeline that can `.transform()` text.
+
+    Examples:
+
+    Scikit-Learn is very flexible and it allows for many representations. 
+    A relatively simple pipeline is shown below. 
+
+    ```python
+    from sklearn.pipeline import make_pipeline
+    from sklearn.decomposition import TruncatedSVD
+    from sklearn.feature_extraction.text import TfidfVectorizer
+    
+    from bertopic.backend import SklearnEmbedder
+
+    pipe = make_pipeline(
+        TfidfVectorizer(),
+        TruncatedSVD(100)
+    )
+
+    sklearn_embedder = SklearnEmbedder(pipe)
+    topic_model = BERTopic(embedding_model=sklearn_embedder)
+    ```
+
+    This pipeline first constructs a sparse representation based on TF/idf and then
+    makes it dense by applying SVD. Alternatively, you might also construct something 
+    more elaborate. As long as you construct a scikit-learn compatible pipeline, you
+    should be able to pass it to Bertopic. 
+
+    !!! Warning 
+        One caveat to be aware of is that scikit-learns base `Pipeline` class does not
+        support the `.partial_fit()`-API. If you have a pipeline that theoretically should
+        be able to support online learning then you might want to explore
+        the [scikit-partial](https://github.com/koaning/scikit-partial) project. 
+    """
+    def __init__(self, pipe):
+        super().__init__()
+        self.pipe = pipe
+
+    def embed(self, documents, verbose=False):
+        """ Embed a list of n documents/words into an n-dimensional
+        matrix of embeddings
+
+        Arguments:
+            documents: A list of documents or words to be embedded
+            verbose: Controls the verbosity of the process
+
+        Returns:
+            Document/words embeddings with shape (n, m) with `n` documents/words
+            that each have an embeddings size of `m`
+        """
+        try: 
+            check_is_fitted(self.pipe)
+            embeddings = self.pipe.transform(documents)
+        except NotFittedError:
+            embeddings = self.pipe.fit_transform(documents)
+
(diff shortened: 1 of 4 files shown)


koaning:
This is a bit of a WIP because there are a few questions on my end. 

- I didn't know which tests you'd like me to add for this. 
- I noticed that I had to install `pip install mkdocstrings-python` in order to run `mkdocs serve`. This is possibly because you didn't hard-pin the `mkdocstrings>=0.8.0` dependency in `setup.py`. 
- I could be wrong, but are you running unit-tests on your docs to make sure that they run? If not, I could explore if it makes sense to run [mktestdocs](https://calmcode.io/labs/mktestdocs.html) there.

MaartenGr:
Thanks for the PR and apologies for the late reply! 

> I didn't know which tests you'd like me to add for this.

This is something I am still trying to figure out seeing as most language backends are quite time-consuming to test from a computational perspective. Especially when using the default pipeline using HDBSCAN, there need to be quite some documents (a couple of thousand) for HDBSCAN to find a sufficient number of topics (> 2) to properly test the topic model. As a result, creating a minimal example of a BERTopic model out of the box can take some time. This, now that I think about it, ties nicely in with the possibility of a more light-weight pipeline and thereby this PR. 

In other words, as much as I would love tests for this. I am afraid it might take you quite some time to figure out the best approach for this. If you have time, please do, but do note that it is by no means expected here. 

> I noticed that I had to install `pip install mkdocstrings-python` in order to run `mkdocs serve`. This is possibly because you didn't hard-pin the `mkdocstrings>=0.8.0` dependency in `setup.py`. Want me to change that as well? 

Nice catch, that would be great! 

> I could be wrong, but are you running unit-tests on your docs to make sure that they run? If not, I could explore if it makes sense to run [mktestdocs](https://calmcode.io/labs/mktestdocs.html) there. 

Nope, I have no unit-tests on the documentation/docstrings themselves. In the past, there definitely have been mistakes in there that would have been corrected with `mktestdocs`. However, the default pipeline with HDBSCAN requires quite a number of documents which I think would result in tests that can run for several hours. It might be worthwhile to explore this a bit in the future but there are quite a few examples that then would have to be re-written like: 

https://github.com/MaartenGr/BERTopic/blob/09c1732997f838050c263ad00ad3c9474e816863/bertopic/_bertopic.py#L828-L833

> I mentioned [scikit-partial](https://github.com/koaning/scikit-partial) in the docs. This is a personal project, so I just want to make sure you're cool with that. 

No problem! It fits nicely with the online topic modeling approach. 

> I kept `verbose=False` in the `embed` method even though it's a no-op. Should that be documented explicitly? 

Seeing as you can define the verbosity in the sklearn pipeline, it might be worthwhile to document that explicitly.

koaning:
Another alternative is to have a weekly job that runs the heavy documentation tests. The thinking here is that you'd like to fix them when they occur, but checking that once a week might be fine. 

I'll leave it up to you to consider if that'll work, but it's a pattern that works grand for other things as well. At scikit-lego, we run a weekly job that checks what happens when we install the latest version of sklearn. It typically catches a breaking bug before our users do.

koaning:
Cool. I think I just addressed your comments. Let me know if I forgot something!

MaartenGr:
> Another alternative is to have a weekly job that runs the heavy documentation tests. The thinking here is that you'd like to fix them when they occur, but checking that once a week might be fine.

> I'll leave it up to you to consider if that'll work, but it's a pattern that works grand for other things as well. At scikit-lego, we run a weekly job that checks what happens when we install the latest version of sklearn. It typically catches a breaking bug before our users do.

Thanks! It indeed might make more sense to leave the heavy testing for a cronjob instead of having to go through that each time something small is tested. I'll take a look at the implementation at scikit-lego. 

> Cool. I think I just addressed your comments. Let me know if I forgot something!

Yep, you definitely did, thanks! There is one last thing that I forgot to mention. In order to minimize the code necessary to use different language models, you can supply, for example, a SentenceTransformer model directly into `embedding_model` without the need to access `bertopic.backend` as it will automatically be recognized as a SentenceTransformerBackend. This is defined in in [backend/_utils.py](https://github.com/MaartenGr/BERTopic/blob/master/bertopic/backend/_utils.py). 

Could you add the SklearnEmbedder also there? As a result, you would be able to do the following instead:

```python
from sklearn.pipeline import make_pipeline
from sklearn.decomposition import TruncatedSVD
from sklearn.feature_extraction.text import TfidfVectorizer

pipe = make_pipeline(
    TfidfVectorizer(),
    TruncatedSVD(100)
)

# Use `pipe` directly instead of needing to import it from BERTopic
topic_model = BERTopic(embedding_model=pipe)
```

koaning:
Ah, nice feature. I'll give that another look and will also check if the docs still make sense.

koaning:
I figured I'd try working on this but hit this error: 

```
Traceback (most recent call last):
  File "/workspaces/BERTopic/check.py", line 5, in <module>
    from bertopic import BERTopic
  File "/workspaces/BERTopic/bertopic/__init__.py", line 1, in <module>
    from bertopic._bertopic import BERTopic
  File "/workspaces/BERTopic/bertopic/_bertopic.py", line 23, in <module>
    import hdbscan
  File "/usr/local/python/3.10.4/lib/python3.10/site-packages/hdbscan/__init__.py", line 1, in <module>
    from .hdbscan_ import HDBSCAN, hdbscan
  File "/usr/local/python/3.10.4/lib/python3.10/site-packages/hdbscan/hdbscan_.py", line 509, in <module>
    memory=Memory(cachedir=None, verbose=0),
TypeError: Memory.__init__() got an unexpected keyword argument 'cachedir'
```

Just to check, this is on your radar?

koaning:
@MaartenGr I made the changes, no tests still, but I think the changes are in.

MaartenGr:
> Just to check, this is on your radar?

Yes, thanks for mentioning though. It is an issue with HDBSCAN that has not taken into account the upcoming changes of joblib in its 1.2.0 release. Although there is a fix for that in HDBSCAN's main branch, it is not yet released into pypi. A quick fix for this in BERTopic would mean fixing joblib to 1.1.0 but might not be preferred due to: https://nvd.nist.gov/vuln/detail/CVE-2022-21797. I don't think it should be an issue in this context but I want to be sure and wait for a HDBSCAN release. For now, removing HDBSCAN and re-installing from its main branch should solve your issue. Would be nice if I had mentioned this to you though before you started working on it, sorry! 😅

All in all, a bit of a tricky situation seeing as pypi does not allow requirements directly from main branches/commits but only from releases. 

> @MaartenGr I made the changes, no tests still, but I think the changes are in.

Thanks for the work, LGTM! I'll wait a few days to see if HDBSCAN gets updated but if it does not, I'll go ahead and merge this.

PR #887, opened 2022-12-19
added diversity parameter to update_topics() function

anubhabdaserrr:
The diversity parameter is a component of topic representation; rather than topic formation (docs -> embeddings -> dimensionality_reduction -> clustering). The update_topics() function should provide the functionality to change the diversity parameter, as and when required. Thanks!

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -990,6 +990,8 @@ class BERTopic:
                       docs: List[str],
                       topics: List[int] = None,
                       n_gram_range: Tuple[int, int] = None,
+                      diversity: float = None,
+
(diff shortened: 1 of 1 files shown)


MaartenGr:
Nice catch, that should definitely be changed! Could you remove the additional whitespace after the diversity parameter and docstrings? Other than that it looks good to me.

anubhabdaserrr:
done!

MaartenGr:
Thanks for the quick update! LGTM, I'll go ahead and merge this now 😄

anubhabdaserrr:
Sorry, I missed out on another parameter - I think top_n_words should also be added, right?

PR #895, opened 2022-12-21
Update _bertopic.py

nbalepur:
Adding functionality in `topics_over_time` to allow users to specify how many terms under each topic they want to see at each timestep *t*. Right now, the default value is 5, but there are use cases where the user may want to see more than 5


MaartenGr:
Thank you for the PR. The difficulty with opening up the `top_n_words_per_time` variable here is that it then should be done for every other variation of BERTopic that generates top n words. This makes the combination with the original `top_n_words` a bit more complex and less intuitive especially when it needs to be implemented for many variations.

PR #968, opened 2023-01-27
Custom document_ids parameter in BERTopic 

dkirman-re:
Inspired by the Top2Vec library, in some cases users may want to provide their own document ids instead of the default value range.
This PR also includes basic type validation


MaartenGr:
Thanks for creating this PR. I typically want to prevent changes to the API as much as possible as tuning and using the API becomes quickly quite complex. Moreover, since the order of the documents is not changed in BERTopic, I believe there is actually no need to track that kind of metadata as that is typically done at the source of the data.

dkirman-re:
I understand, the prompt feedback is much appreciated!

PR #981, opened 2023-02-03
fix 'cuml' w/ 'hdbscan' type checks in `cluster/_utils.py`

FelSiq:
I believe that current type checks in `bertopic/cluster/_utils.py` for `cuml` + `hdbscan` clusterers are incorrect.

Since the "cuml" string is always evaluating to True (see lines 22, 34, and 45 in `bertopic/cluster/_utils.py`), any subclass of HDBSCAN is triggering the `cuml` import, raising an `ImportError` when not installed. Even if installed, calling `cuml_hdbscan` module may not be adequate for all HDBSCAN subclasses.

I'm opening this PR with a simple solution proposal, but feel free to find another way to fix this issue.


MaartenGr:
Thanks for the PR! This indeed should resolve any future HDBSCAN-like models to be implemented as well as custom models that do not yet support probability calculation.

PR #1002, opened 2023-02-14
Fix typo in textgenerator - [DOCUMENTS] mapping

dkopljar27:
There is a typo in the current version of the code (line 98):
 - _create_prompt is expecting [DOCUMENTS] while topic_model._extract_representative_docs looks for [DOCUMENT]


MaartenGr:
Nice catch, thanks!

PR #1073, opened 2023-03-06
Inconsistent naming - get_topic_labels to generate_topic_labels

rolanderdei:
There has been an inconsistent naming in the documentation between get_topic_labels and generate_topic_labels. All get_topic_labels mentions have been changed to the correct one. (generate_topic_labels)


MaartenGr:
Nice catch, thanks for the PR!

Issue #1077, opened 2023-03-07
Suggestion: Change scaling of the max_distance levels in visualize_hierarchical_documents


zilch42:
Hi Maarten,

The `visualize_hierarchical_documents` function is super cool for visualising the topic structure so thank you for your work on that. 

As a suggestion, I wonder if it would be helpful to have some control over the resolution at different ends of the hierarchy? Once the number of topics becomes reasonably large (>20ish) the plot becomes less helpful because the labels dominate and the colours get reused repeatedly. With very large numbers of topics (>100), even with a higher `nr_levels`, most of those levels aren't very useful because they are just a soup of hundreds of topics. What would be great is having less resolution at the lower end of the slider and more resolution at the upper end of the slider where there are fewer higher-level topics, as it's quite interesting to see how those higher-level topics split out and merge. 

![image](https://user-images.githubusercontent.com/39004650/223337564-821390b7-9d89-442c-9463-7c757a1a2ebf.png)

Splitting the distance indices over a logarithmic scale rather than a linear one seems to produce something more useful to me.

I'm sure there is a more elegant way to do this but as an example, replacing this line with the code below produced something more usable for me. 
https://github.com/MaartenGr/BERTopic/blob/d665d3f8d8c7c1736dc82b1df8839ced56a2adb6/bertopic/plotting/_hierarchical_documents.py#L150

```
    log_indices = np.round(np.logspace(start=math.log(1,10), stop=math.log(len(distances)-1,10), num=nr_levels)).astype(int).tolist()
    log_indices.reverse()
    max_distances = [distances[i] for i in log_indices]
```
Ideally the user should probably have some control over the scaling too.

Open to your thoughts.

MaartenGr:
Thanks for the suggestion! Could you perhaps create a reproducible example of this? Or perhaps show the effect of what you propose? In practice, I wonder what the effect would be of your suggestion on use cases where hierarchical topic modeling is applied on < 30 topics with > 5 levels.

zilch42:
Sure thing. Forgive me, I am not 100% sure on the best way to provide a reproducible example that relies on changes to your source code. Is a pull request the easiest way to do that? Anyway, the following example assumes the following changes have been made to BERTopic:

`log_scale_levels: bool = False` added as an input argument in 2 places :
[...]

PR #1106, opened 2023-03-20
added level_scale argument to visualize_hierarchical_documents

zilch42:
Adds `level_scale` argument to `visualize_hierarchical_documents`. Can take the following values: 'lin', 'linear', 'log', 'logarithmic'. 

Logarithmic options apply a logarithmic scale to the distance vector allowing for greater resolution of topic merging at higher levels which is useful when the number of topics is high. 

Addresses issue #1077

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -2152,6 +2152,7 @@ class BERTopic:
                                          hide_annotations: bool = False,
                                          hide_document_hover: bool = True,
                                          nr_levels: int = 10,
+                                         level_scale: str = 'linear', 
                                          custom_labels: bool = False,
                                          title: str = "<b>Hierarchical Documents and Topics</b>",
                                          width: int = 1200,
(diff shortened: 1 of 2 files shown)


MaartenGr:
I'm not entirely sure why the pipeline is currently failing. I would have to look into that a bit more. In part, I believe it is a result of not having created sufficient clusters.

zilch42:
Thanks Maarten, 

I'm not heaps familiar with how to read those build logs, but it looks like it is failing in a whole range of places that aren't related to the things that have changed here? Is the same pipeline run against the master branch?

I did notice when I was putting together my case in response to your initial questions that `visualize_hierarchical_documents` failed when `nr_topics=5`, but it seemed to fail both with and without my modifications so I didn't think it was something on my end... (I'm not sure why anyone would need to look into hierarchial documents with so few topics :) ) 

But anyway, if there is anything I need to go back and look at, please let me know.

MaartenGr:
> But anyway, if there is anything I need to go back and look at, please let me know.

Thanks! I just fixed the pipeline (there were some API changes in plotly and some issues with the topic generation from few documents) but it should be working now. Based on the changes, I'll go ahead and merge this now. Thanks for the PR!

PR #1139, opened 2023-03-30
Update usecases.md

abdu355:
Add SnowHeap's use case


MaartenGr:
Thank you for taking the time to create the PR and write down the use case! I have to be honest, it does not currently read like a specific use case of BERTopic. It seems to focus mostly on the excellence of the organization you mentioned. Albeit important, the page is intended for specific use cases. Shout-outs to other organizations are welcome but not their main purpose.

PR #1173, opened 2023-04-08
fix hierarchy viz and handle any form of distance matrix

### LINKED ISSUE #923: visualize_hierarchy mix up of labels / hover over
I've found that for some datasets I'm getting different hover over text from the labels when using the visualize_hierarchy function. Tracking up the tree, the hover over labels seem to be in the correct order but the axis labels don't match at the lowest level. I've tried both with and without updating the topic names and this does not appear to make a difference. I think that the two lists are somehow being ordered differently but I can't track down where this occurs.

### LINKED ISSUE #1021: Hierarchical Visualization not lining up
```
from bertopic import BERTopic
from scipy.cluster import hierarchy as sch

docs = fetch_20newsgroups(subset='all',  remove=('headers', 'footers', 'quotes'))["data"]
topic_model = BERTopic(verbose=True)
topics, probs = topic_model.fit_transform(docs)

linkage_function = lambda x: sch.linkage(x, 'single', optimal_ordering=True)
hierarchical_topics = topic_model.hierarchical_topics(docs, linkage_function=linkage_function)

fig = topic_model.visualize_hierarchy(hierarchical_topics=hierarchical_topics)

fig.show()
```

image

Tested this using my own data and noticed the topic names that show up in the hoverbox dont match the label. I switched to the newsgroup documents to check and it is doing the same thing. It also seems like things arent lining up correctly in terms of merged topic names. In my example, two end topics that both involve carbon have a shared parent topic that ends up being something totally different.

Any ideas whats going on here?

### LINKED ISSUE #1063: visualize_hierarchy has new merged topics out of order
Hi Maarten,

I think I have encountered another topic ordering issue though I don't think it is directly related to #1024 (I do have an outlier cluster in this case and the issue appears to be present both in v0.14.0 and v0.14.1 so it wasn't introduced by the recent fix for #1024 - thankyou for that btw)

The descriptions of the merged topics displayed on hover from `visualize_hierarchy()` often don't make any sense with respect to the topics they are apparently merging, and also differ from the hierarchy presented in `get_topic_tree()`. 

Based on the code below, the merged topic from 191 and 2 should be 'test_hello_whatta_ites_cheek' according to `get_topic_tree`, but in `visualize_hierarchy` that topic is 'weaver_trail_spence_cooper_court'.

```
from bertopic import BERTopic
import numpy as np

data = fetch_20newsgroups(subset='all',  remove=('headers', 'footers', 'quotes'))
docs = data['data']

embedding_model = SentenceTransformer("all-MiniLM-L6-v2")
embeddings = embedding_model.encode(docs, show_progress_bar=True)

topic_model = BERTopic(embedding_model=embedding_model)
topics, probs = topic_model.fit_transform(docs, embeddings)

hierarchical_topics = topic_model.hierarchical_topics(docs)

topic_model.visualize_hierarchy(hierarchical_topics=hierarchical_topics)
```

Hierarchy

```
tree = topic_model.get_topic_tree(hierarchical_topics)
print(tree)
```
Tree

I believe it is an alignment issue in `visualize_hierarchy`. The merged topics in `get_topic_tree` look really good to me. They make a lot of sense. It's just the merged topics on the visualisation that don't seem to line up properly.

elashrry:
This fixes #923, fixes #1063 , and fixes #1021 which are just duplicates of the first one. 

I added a utility function to make sure the distance matrix computed by a given distance_function is a condensed 1-D array. If not, the function convert it and return the result. It can be used directly as I did in [`_bertopic.pyL877`](https://github.com/elashrry/BERTopic/blob/fix-hierarchy-viz/bertopic/_bertopic.py#L877):
```
X = validate_distance_matrix(X, embeddings.shape[0])
```

or to wrap around a function as I did in [`_hierarchy.py#L132-L133`](https://github.com/elashrry/BERTopic/blob/fix-hierarchy-viz/bertopic/plotting/_hierarchy.py#L132-L133):

```
distance_function_viz = lambda x: validate_distance_matrix(
        distance_function(x), embeddings.shape[0])
```

I also added this information in the docstring to let the user know what is possible. I didn't find things to changes in the tutorials. 

I added a testing notebook to test the code provided in the mentioned issues. (to be deleted after reviewing the PR)

Let me know if you have any comments!


bertopic/_utils.py
```diff
@@ -92,3 +93,54 @@ def __getattr__(self, *args, **kwargs):
...
+            raise ValueError("The condensed distance matrix must have "
+                                "shape (n*(n-1)/2,).")
+    elif len(s) == 2:
+        # check it is square
+        if s[0] != s[1]:
+            raise ValueError("The distance matrix must be square.")
+        # check it has correct size
+        if s[0] != n_samples:
+            raise ValueError("The distance matrix must be of shape "
+                                "(n, n) where n is the number of documents.")
```
MaartenGr, line 131: I think these two if statements can be captured in a single statement since the last ValueError essentially also states that the distance matrix should be square.

bertopic/_utils.py
```diff
@@ -92,3 +93,54 @@ def __getattr__(self, *args, **kwargs):
...
+        # force zero diagonal and convert to condensed
+        np.fill_diagonal(X, 0)
+        X = squareform(X)
+    else:
+        raise ValueError("The distance matrix must be either a 1-D condensed "
+                            "distance matrix of shape (n*(n-1)/2,) or a "
+                            "2-D square distance matrix of shape (n, n)."
+                            "where n is the number of documents."
+                            "Got a distance matrix of shape %s" % str(s))
```
MaartenGr, line 140:
```suggestion
        raise ValueError("The distance matrix must be either a 1-D condensed "
                         "distance matrix of shape (n*(n-1)/2,) or a "
                         "2-D square distance matrix of shape (n, n)."
                         "where n is the number of documents."
                         "Got a distance matrix of shape %s" % str(s))
```

Could you remove whitespace?

bertopic/plotting/_hierarchy.py
```diff
@@ -9,6 +9,7 @@
 import plotly.graph_objects as go
 import plotly.figure_factory as ff
 
+from .._utils import validate_distance_matrix
```
MaartenGr, line 12:
```suggestion
from bertopic._utils import validate_distance_matrix
```
Please import it as above.

bertopic/_utils.py
MaartenGr:
```suggestion
        if s[0] != s[1]:
            raise ValueError("The distance matrix must be square.")
        # check it has correct size
        if s[0] != n_samples:
            raise ValueError("The distance matrix must be of shape "
                             "(n, n) where n is the number of documents.")
```
Whitespaces

bertopic/_utils.py
```diff
@@ -92,3 +93,54 @@ def __getattr__(self, *args, **kwargs):
...
+
+    # Make sure its entries are non-negative
+    if np.any(X < 0):
+        raise ValueError("Distance matrix cannot contain negative values.")
```
MaartenGr, line 144: I am not sure whether this is necessary as this issue is handled and raised within `Scipy` already. Perhaps remove?

MaartenGr (changes requested):
Thanks for the PR and the extensive docs, greatly appreciated! A few minor things to change before it will be merged (after CI has passed).

bertopic/_utils.py
elashrry: the `squareform` from SciPy doesn't raise error if the entries are non-negative. 

```
X = np.array([1, -2, 3,])
squareform(X)
```
Output: 
```
array([[ 0,  1, -2],
       [ 1,  0,  3],
       [-2,  3,  0]])
```

elashrry:
I just pushed the changes you requested. I don't know why the test is failing. Let me know if I can help!

bertopic/_utils.py
MaartenGr: That's true but the linkage function will give the error but it's alright to keep it here.

MaartenGr:
The workflow is failing due to an issue with the testing pipeline. It automatically pulls pandas 2.0 which has a slight change that breaks some specific tests, unfortunately. Having said that, if it runs successfully locally for you, then I think it would be alright to merge it.

elashrry:
Great, it run successfully on my computer.

MaartenGr:
Awesome, thanks for the work on this PR! It is highly appreciated.

elashrry:
You are welcome! and thank you for this nice library!

PR #1205, opened 2023-04-22
Efficient, faster, and safer serialization with HuggingFace

MaartenGr:
# Efficient Saving with HuggingFace

Through a great collaboration with HuggingFace and the [BERTransfer](https://github.com/opinionscience/BERTransfer) initiative, a set of features aimed at efficiently saving, loading, and using BERTopic.  

The underlying idea is that the modularity of BERTopic allows it to remove the dimensionality and clustering models **after** fitting BERTopic as transforming unseen documents to topics can be done by comparing document embeddings with topic embeddings. Moreover, many embedding models are already on the HuggingFace Hub, so only pointers are necessary to be saved. 

As a result, we can compress a BERTopic model from ~500MB to only ~5MB by only needing to save the following things:
* Pointer to the embedding model (e.g., "sentence-transformers/all-MiniLM-L6-v2")
* Topic representations (topics, topic labels, custom labels, topic sizes, etc.)
* BERTopic hyperparameters
* c-TF-IDF matrix

All of the above can be saved quite efficiently and is all that is necessary to reproduce a BERTopic model. As a result, there are a bunch of new functionalities and advantages with BERTopic, especially from a production setting:

* Nearly **all functionality** of BERTopic is kept
* Create **small** BERTopic models
   * Training on millions of documents would only result in a model of around 20MB-50MB!
* **Easily share** models as they are quite small
  * Sharing with **HuggingFace Hub** is now also possible 
* **Backwards compatible** from this version onwards
* More easily used in **production**
* Save BERTopic **safely** with **safetensors**
> **NOTE:** You can find an example of a pre-trained BERTopic model [here](https://huggingface.co/MaartenGr/BERTopic_ArXiv)

## HuggingFace Hub

```python
from bertopic import BERTopic

# Train model
topic_model = BERTopic().fit(my_docs)

# Push to HuggingFace Hub
topic_model.push_to_hf_hub(
    repo_id="MaartenGr/BERTopic_ArXiv",
    save_ctfidf=True,
    save_embedding_model="sentence-transformers/all-MiniLM-L6-v2"
)

# Load from HuggingFace
loaded_model = BERTopic.load("MaartenGr/BERTopic_ArXiv")
```

## Local Saving

```python
from bertopic import BERTopic

# Train model
topic_model = BERTopic().fit(my_docs)

# Save locally as safetensor with c-TF-IDF and a pointer to an embedding model
embedding_model = "sentence-transformers/all-MiniLM-L6-v2"
topic_model.save("model_dir", serialization="safetensors", save_ctfidf=True, save_embedding_model=embedding_model)

# Load from local dir
loaded_model = BERTopic.load("model_dir")
```

### To do
* Update topic embeddings to be centroids as they are typically more accurate for inference (less so for representation purposes)
  * Additionally, since c-TF-IDF can capture domain-specific data more easily, also use c-TF-IDF to perform inference (or give the option to do so) 
  * The old topic embeddings, weighted average of c-TF-IDF word embeddings, can perhaps still be kept for representation purposes (e.g., comparing topic representations)

### DIFF
--- a/bertopic/_save_utils.py
+++ b/bertopic/_save_utils.py
@@ -1,0 +1,380 @@
+import os
+import sys
+import json
+import numpy as np
+import scipy.sparse as sp
+
+from pathlib import Path
+from scipy.sparse import csr_matrix
+from tempfile import TemporaryDirectory
+from sklearn.feature_extraction.text import CountVectorizer
+
+from bertopic import BERTopic
+from bertopic.cluster import BaseCluster
+from bertopic.backend import BaseEmbedder
+from bertopic._bertopic import TopicMapper
+from bertopic.backend._utils import select_backend
+from bertopic.dimensionality import BaseDimensionalityReduction
+
+# HuggingFace Hub
+try:
+    from huggingface_hub import (
+        create_repo, get_hf_file_metadata,
+        hf_hub_download, hf_hub_url,
+        repo_type_and_id_from_hf_id, upload_folder)
+    _has_hf_hub = True
+except ImportError:
+    _has_hf_hub = False
+
+# Typing
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+from typing import Union, Mapping, Any
+
+# Safetensors check
+try:
+    import safetensors.torch
+    _has_safetensors = True
+except ImportError:
+    _has_safetensors = False
+
+# Pytorch check
+try:
+    import torch
+    _has_torch = True
+except ImportError:
+    _has_torch = False
+
+TOPICS_NAME = "topics.json"
+CONFIG_NAME = "config.json"
+
+HF_WEIGHTS_NAME = "topic_embeddings.bin"  # default pytorch pkl
+HF_SAFE_WEIGHTS_NAME = "topic_embeddings.safetensors"  # safetensors version
+
+CTFIDF_WEIGHTS_NAME = "ctfidf.bin"  # default pytorch pkl
+CTFIDF_SAFE_WEIGHTS_NAME = "ctfidf.safetensors"  # safetensors version
+CTFIDF_CFG_NAME = "ctfidf_config.json"
+
+
+def push_to_hf_hub(
+        model,
+        repo_id: str,
+        commit_message: str = 'Add BERTopic model',
+        token: str = None,
+        revision: str = None,
+        private: bool = False,
+        create_pr: bool = False,
+        model_card: Mapping[str, Any] = None,
+        serialization: str = "safetensors",
+        save_embedding_model: str = None,
+        save_ctfidf: bool = False,
+        ):
+    """ Push your BERTopic model to a HuggingFace Hub
+
+    Arguments:
+        repo_id: The name of your HuggingFace repository
+        commit_message: A commit message
+        token: Token to add if not already logged in
+        revision: Repository revision
+        private: Whether to create a private repository
+        create_pr: Whether to upload the model as a Pull Request
+        model_card: Create a model card when creating the repository
+        serialization: The type of serialization.
+                       Either `safetensors` or `pytorch`
+        save_embedding_model: A pointer towards a HuggingFace model to be loaded in with
+                                SentenceTransformers. E.g.,
+                                `sentence-transformers/all-MiniLM-L6-v2`
+        save_ctfidf: Whether to save c-TF-IDF information
+    """
+
+    # Create repo if it doesn't exist yet and infer complete repo_id
+    repo_url = create_repo(repo_id, token=token, private=private, exist_ok=True)
+    _, repo_owner, repo_name = repo_type_and_id_from_hf_id(repo_url)
+    repo_id = f"{repo_owner}/{repo_name}"
+
+    # Temporarily save model and push to HF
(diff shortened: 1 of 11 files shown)


bertopic/_save_utils.py
```diff
@@ -0,0 +1,315 @@
...
+        except:
+            ctfidf_tensors = hf_hub_download(path, CTFIDF_WEIGHTS_NAME, revision=None)
+            ctfidf_tensors = torch.load(ctfidf_tensors, map_location="cpu")
+    except:
+        ctfidf_config, ctfidf_tensors = None, None
+
+    return topics, params, tensors, ctfidf_tensors, ctfidf_config
+
+
+def generate_readme(model_card: dict, model_name: str):
```
davanstrien, line 172: For this, it might be worth exploring https://huggingface.co/docs/huggingface_hub/guides/model-cards at some point. There are a few possible enhancements we could make there:

- Trying to autogenerate some parts of the model card (for example, the Transformers Trainer API adds hyperparameters to a model card https://huggingface.co/biglam/detr-resnet-50_fine_tuned_nls_chapbooks#training-hyperparameters. I could imagine that things like the number of topics, number of training docs etc., could all be of interest. 

- Creating a template model card, i.e. something like (https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/templates/modelcard_template.md). This can be helpful as a starting point for a model card structure even if most sections are not completed. 

I think this can all come later, and happy to help with this part if you think it might interest you.

davanstrien (commented):
This is looking great! Have played around with this locally, and this seems to be working nicely.

bertopic/_save_utils.py
```diff
@@ -0,0 +1,315 @@
...
+
+# Typing
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+from typing import Union, Mapping, Any
+
+# Safetensors check
+try:
```
davanstrien, line 28: Would it be worth doing this check inside the save function? For the user who is running in Colab or another notebook environment, they might hit the assertion error about safetensors when they try and save a model, but since the import check is done earlier, they would need to reload the module even after they pip install safetensors.

bertopic/_save_utils.py
MaartenGr: Oh nice, that's perfect! That would make the process rather straightforward. Just to be sure I got this correctly. I would only need to:
* Create a template
* Create a model card from that template
* Extract the text from the model card
* and finally write that to `.md` the same way I did before. 

I want to make sure that when I run `upload_folder`, the model card is not a separate commit. 

All in all, I do not think it would be too much work, so I think I will simply add it to this PR. That way, there is little left for saving/loading a BERTopic model.

bertopic/_save_utils.py
MaartenGr: Thanks, will definitely do this. It is indeed something users will run into since `safetensors` is not (yet) part of the transformers base requirements.

bertopic/_save_utils.py
MaartenGr: Ah, it seems that to use `.from_template` an installation of `Jinja2` is a requirement whilst it is not a requirement of BERTopic.  Instead, I will likely use f-strings, albeit a bit hackier, to generate the template.

davanstrien: > Ah, it seems that to use .from_template an installation of Jinja2 is a requirement whilst it is not a requirement of BERTopic. Instead, I will likely use f-strings, albeit a bit hackier, to generate the template.

Sorry for the slow reply, yes I think in this case to avoid a dependency f-strings probably make the most sense. Let me know if you need any help/input with this.

MaartenGr: Great, I finished up the last changes. I am currently working on support for both multi-modal (cluster images and/or text) and multi-aspect topic modeling (multiple representations simultaneously, keywords, summaries, sentiment, etc.). I intend to also save those aspects but since they are mostly just values/text that can be put in the `topics.json`, that process should be straightforward. 

So I believe the best thing to do may be to merge this PR and then continue with the multi-modal/multi-aspect topic modeling before slightly updating the saving procedure.

davanstrien: That sounds awesome! Super cool to be adding multi-modal support for this (it gives me a few more use cases to use this!!). 

Let me know if you want me to review any of these changes or input anywhere else.

MaartenGr: I'll go ahead and merge this now so that I can continue with the multi-modal support

> Let me know if you want me to review any of these changes or input anywhere else.

Great, thanks! Most likely, I will ping you when I start saving the multi-modal model as I am not quite sure what the best way to save those images is, but I will let you know by then when I have something you can look at.

davanstrien: Thanks, @MaartenGr, and really excited about all of this work!

vantubbe:
This is an awesome addition. I was able to create and save a model on a linux distro and then load successfully on windows. Great work!!!!

MaartenGr:
@vantubbe Thanks! Glad to hear that it is a worthwhile feature to add. There are going to be a few more changes with respect to multi-modal and multi-aspect topic modeling before this is fully integrated in the HF Hub but hopefully it will only take a couple of weeks.

PR #1232, opened 2023-05-04
Multimodal and Multi-Aspect Topic Modeling

MaartenGr:
# **Features**
* **Multimodal** Topic Modeling
  * Topic modeling on images only
  * Topic modeling on images + text
* **Multi-Aspect** Topic Modeling
  * Create multiple topic representations simultaneously 
* Updated **topic embeddings**
  * Centroids by default and c-TF-IDF weighted embeddings for `partial_fit` and `.update_topics`
  * Centroids work typically better for inference
* Use `custom_labels="Aspect1"` to use the aspect labels for visualizations instead
* Added example of [summarizing topics](https://maartengr.github.io/BERTopic/getting_started/representation/representation.html#summarization) with OpenAI's GPT-models

# **Multimodal Topic Modeling**

Often, documents or text are accompanied by imagery or the other way around. For example, social media images with captions and products with descriptions. Topic modeling has traditionally focused on creating topics from textual representations. However, as more multimodal representations are created, the need for multimodal topics increases. 

Therefore, **multimodal topic modeling** is introduced! During the `fit` or `fit_transform` stages, you can now create image representations of each single topic.

## **Text + Images**

To do so, we first have to collect our images with captions:

```python

ds = load_dataset("maderix/flickr_bw_rgb")
images = ds["train"]["image"]
docs = ds["train"]["caption"]
```

Then, we can train our multimodal topic model with BERTopic and view the resulting images per topic:

```python
from bertopic.representation import VisualRepresentation

# Additional ways of representing a topic
visual_model = VisualRepresentation()

# Add all models together to be run in a single `fit`
representation_model = {
   "Visual_Aspect":  visual_model,
}
topic_model = BERTopic(representation_model=representation_model, verbose=True, min_topic_size=20)
```

If you want an overview of the topic images together with their textual representations in `jupyter`, you can run the following:

```python
from IPython.display import HTML
from io import BytesIO
import base64

def image_base64(im):
    if isinstance(im, str):
        im = get_thumbnail(im)
    with BytesIO() as buffer:
        im.save(buffer, 'jpeg')
        return base64.b64encode(buffer.getvalue()).decode()

def image_formatter(im):
    return f'<img src="data:image/jpeg;base64,{image_base64(im)}">'

df = topic_model.get_topic_info().drop("Representative_Docs", 1).drop("Name", 1)
HTML(df.to_html(formatters={'Visual_Aspect': image_formatter}, escape=False))
```

It seems to do quite well, here are some examples:

image

image

image

## **Images Only**

There might be times when you only have images as your data without any textual data. Normally, you would not immediately look towards topic modeling as that is known for a method working with textual data. However, we the

```python
from bertopic import BERTopic
from bertopic.backend import MultiModalBackend
from bertopic.representation import KeyBERTInspired, VisualRepresentation

# Image embedding model
embedding_model = MultiModalBackend('clip-ViT-B-32', batch_size=32)

# Image to text representation model
representation_model = {
    "Visual_Aspect": VisualRepresentation(text_to_image_model="nlpconnect/vit-gpt2-image-captioning")
}

# Train our model with images only
topic_model = BERTopic(representation_model=representation_model, embedding_model=embedding_model)
topics, probs = topic_model.fit_transform(documents=None, images=images)
```

Even with images only, we can generate amazing results:

image

image

image

## Multimodal backend

There is now also a new multimodal backend that relies on `sentence-transformers`:

```python
from bertopic.backend import MultiModalBackend
model = MultiModalBackend('clip-ViT-B-32', batch_size=32)

# Embed documents only
doc_embeddings = model.embed_documents(docs)

# Embeding images only
image_embeddings = model.embed_images(images)

# Embed both images and documents, then average them
doc_image_embeddings = model.embed(docs, images)
```
# **Multi-Aspect Topic Modeling**

During the development of BERTopic, many different types of representations can be created, from keywords and phrases to summaries and custom labels. There is a variety of techniques that one can choose from to represent a topic. As such, there are a number of interesting and creative ways one can summarize topics. A topic is more than just a single representation. 

Therefore, **multi-aspect topic modeling** is introduced! During the `fit` or `fit_transform` stages, you can now get multiple representations of a single topic:

```python
from bertopic.representation import PartOfSpeech
from bertopic.representation import MaximalMarginalRelevance

# The main representation of a topic
main_representation = KeyBERTInspired()

# Additional ways of representing a topic
aspect_model1 = PartOfSpeech("en_core_web_sm")
aspect_model2 = [KeyBERTInspired(top_n_words=30), MaximalMarginalRelevance(diversity=.5)]

# Add all models together to be run in a single `fit`
representation_model = {
   "Main": main_representation,
   "Aspect1":  aspect_model1,
   "Aspect2":  aspect_model2 
}
```

In the example above, we define the main topic representation using the `"Main"` key and assign the `KeyBERTInspired` model to it. If we want to use the default representation, you can simply omit the `"Main"` key. 

Then, we have 2 additional topic representations or **aspects**:
* A PartOfSpeech model that uses POS tagging to optimize the `"Main"` representation
* A pipeline consisting of KeyBERTInspired and MMR to further optimize the `"Main"` representation

As a result, we get the following table with representations: 

table


bertopic/_bertopic.py
```diff
@@ -3184,12 +3220,15 @@ def _extract_representative_docs(self,
         Returns:
             repr_docs_mappings: A dictionary from topic to representative documents
             representative_docs: A flat list of representative documents
-            repr_doc_indices: The indices of representative documents
+            repr_doc_indices: Orrdere indices of representative documents
```
proselotis, line 3223:
```suggestion
            repr_doc_indices: Ordered indices of representative documents
```

bertopic/backend/_multimodal.py
```diff
@@ -0,0 +1,150 @@
...
+
+class MultiModalBackend(BaseEmbedder):
+    """ Multi-modal backend using Sentence-transformers
+
+    The sentence-transformers embedding model used for 
+    generating word, document, and image embeddings.  
+
+    Arguments:
+        embedding_model: A sentence-transformers embedding model
+        batch_size: The sizes of image batches to pass
```
proselotis, line 19:
```suggestion
        batch_size: The size of image batches to pass
```

bertopic/_bertopic.py
MaartenGr: Thanks for the feedback! This is still all a work-in-progress and is likely to change significantly, especially when new features come in, but I'll make sure to change it accordingly if it is still there after a few commits!

bertopic/backend/_multimodal.py
MaartenGr: See above 😄

Steven-1124:
Hi! I am using Multimodal model and want to sent my own ebeddings to speed up the model; while the error persists: 

**Token indices sequence length is longer than the specified maximum sequence length for this model (78 > 77). Running this sequence through the model will result in indexing errors**

I searched and found that It may be due to the restriction of the model. Do you know how to solve it? Thanks~~

My code: 

`model = MultiModalBackend('clip-ViT-B-32', batch_size=32)
`
`embeddings = model.embed(docs, images)`

MaartenGr:
@Steven-1124 That is a result of the context length of the embedding model. If it is trained on a number of tokens, then it will not accept longer documents and truncation of the documents might be needed.

PR #1287, opened 2023-05-22
Add pipeline_tag to model card 

davanstrien:
The `pipeline_tag` is used on the Hugging Face hub to determine what tasks a model supports. This determines which inference API/widget is used for the model.

### DIFF
--- a/bertopic/_save_utils.py
+++ b/bertopic/_save_utils.py
@@ -54,6 +54,7 @@ MODEL_CARD_TEMPLATE = """
 tags:
 - bertopic
 library_name: bertopic
+pipeline_tag: text-classification
 ---
 
 # {MODEL_NAME}


bertopic/_save_utils.py
```diff
@@ -363,6 +371,13 @@ def save_config(model, path: str, embedding_model):
 
     return config
 
+def check_has_visual_aspect(model):
```
davanstrien, line 374: This is heavily duplicating code in `save`, but I didn't want to take the liberty of refactoring this out!

bertopic/_save_utils.py
```diff
@@ -284,6 +285,13 @@ def generate_readme(model, repo_id: str):
     model_card = model_card.replace("{HYPERPARAMS}", params)
     model_card = model_card.replace("{FRAMEWORKS}", frameworks)
     
+    # Fill Pipeline tag
+    has_visual_aspect = check_has_visual_aspect(model)
+    if not has_visual_aspect:
+        model_card = model_card.replace("{PIPELINE_TAG}", "text-classification")
+    else:
+        model_card = model_card.replace("pipeline_tag: {PIPELINE_TAG} /n","") # TODO add proper tag for this instance
```
davanstrien, line 293: I will aim to find a way of implementing this properly

bertopic/_save_utils.py
MaartenGr: Feel free to refactor stuff but this is fine by me for now!

bertopic/_save_utils.py
MaartenGr: @davanstrien Great, should we save this for later? If so, then I will go ahead and merge this PR.

davanstrien: I think that makes sense for now since this will probably require a bit more work.

Issue #1317, opened 2023-06-02
model.transform() throwing error when using cuml for HDBSCAN with calculate_probabilities=True


slice-pranay:
Hi Maarten

Firstly, thank you for this amazing library. I'm generating topics on newsgroups data for testing and I am using cuML for UMAP and HDBSCAN. I have set the `calculate_probabilites = True` and performed fit_transform() on the data. It worked fine and gave good results. When I try to run transform() on new data it gives an error `AttributeError: 'tuple' object has no attribute 'shape'`. When i set `calculate_probabilities = False` this function works fine.

The libraries i am using are
bertopic==0.15.0
cuml-cu11==23.4.1
cudf-cu11==23.4.1
cuda toolkit 11.8

I am running on a virtual ubuntu machine with Tesla T4 GPU.

The code to reproduce this error

```
from bertopic import BERTopic
from cuml.cluster import HDBSCAN
from cuml.manifold import UMAP
import pandas as pd
import numpy as np
from sklearn.datasets import fetch_20newsgroups
docs = fetch_20newsgroups(subset='all',  remove=('headers', 'footers', 'quotes'))['data']

train = docs[:15000]
test = docs[15000:]

umap_model = UMAP(n_components=5, n_neighbors=10, min_dist=0.0)
hdbscan_model = HDBSCAN(min_samples=25, min_cluster_size=50, gen_min_span_tree=True, prediction_data = True)

topic_model = BERTopic(umap_model=umap_model, hdbscan_model=hdbscan_model, calculate_probabilities=True, verbose=True)
topics,probs = topic_model.fit_transform(train)

topics_test, probs_test = topic_model.transform(test)
```

The error that comes when i run this 
<img width="1002" alt="Screenshot 2023-06-02 at 5 26 26 PM" src="https://github.com/MaartenGr/BERTopic/assets/97085202/97a16841-01f2-459e-9716-0733c657018c">

Can you please guide me in solving this error.

beckernick:
Perhaps this if block may need to be updated to reflect that `approximate_predict` returns a tuple of `(labels, probabilities)` (even if only the probabilities will be returned by the function).

https://github.com/MaartenGr/BERTopic/blob/fca5a4f9df149609c7e3458d6b2c421194cea62c/bertopic/cluster/_utils.py#L52-L56

[...]

PR #1324, opened 2023-06-08
add support for cuml hdbscan membership_vector

stevetracvc:
cuml version 23.04 now has membership_vector, so this allows topic_model.transform() to calculate the probability matrix if using a cuml-based hdbscan model

As I was preparing to submit this PR, I saw #1317 and that some of you have already figured this out :smile: 

I had included batch_size as a new parameter to hdbscan_delegator just in case someone needed to pass in a batch_size somehow

### DIFF
--- a/bertopic/cluster/_utils.py
+++ b/bertopic/cluster/_utils.py
@@ -1,9 +1,11 @@
 import hdbscan
 import numpy as np
 
-        
-def hdbscan_delegator(model, func: str, embeddings: np.ndarray = None):
-    """ Function used to select the HDBSCAN-like model for generating 
+
+def hdbscan_delegator(model, func: str,
+                      embeddings: np.ndarray = None,
+                      batch_size: int = 4096 ):
+    """ Function used to select the HDBSCAN-like model for generating
     predictions and probabilities.
 
     Arguments:
@@ -14,6 +16,7 @@ def hdbscan_delegator(model, func: str, embeddings: np.ndarray = None):
                 - "membership_vector"
         embeddings: Input embeddings for "approximate_predict"
                     and "membership_vector"
+        batch_size: batch_size for cuml hdbscan
     """
 
     # Approximate predict
@@ -42,7 +45,7 @@ def hdbscan_delegator(model, func: str, embeddings: np.ndarray = None):
             return cuml_hdbscan.all_points_membership_vectors(model)
 
         return None
-    
+
     # membership_vector
     if func == "membership_vector":
         if isinstance(model, hdbscan.HDBSCAN):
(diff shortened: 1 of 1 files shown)


MaartenGr:
Thanks for the PR! Did you check whether the `try/except` clause works? According to the issue that you refer `approximate_predict` returns a tuple, so I think it will break if you are using cuML < 23.04. Also, I am wondering whether the `batch_size` parameter should be added since it will be fixed in cuML 23.08.

stevetracvc:
Good catch, that should be an `AttributeError` instead

I'm fine removing the batch_size param, but I added it so you can use a custom batch size if necessary. I think it's important to leave the check in there, because telling people this function won't work for cuml 23.04 or .06 isn't great.

I just left it as it was, and yes, `approximate_predict` returns a tuple. What should be the expected return if someone tries to get membership_vector when using an incompatible cuml?

MaartenGr:
Apologies for the late reply. I am wondering whether it would not be better to skip over handling the `batch_size` parameter since that was fixed in the most recent release. Although I understand that previous versions are still used, I believe most users would use the newest stable release automatically when installing cuML. I'm afraid it will give problems later on when more details about `batch_size` parameter might change or be removed.

HeadCase:
I am currently experiencing Cuda out_of_memory (OOM) errors on my (admittedly large) dataset when `hdbscan_delegator` makes the call to `all_points_membership_vectors`. I know that I can continue to tweak `min_cluster_size`/`min_samples` to reduce memory constraints (or switch off `calculate_probabilities`), but I would prefer those to be last resorts. My understanding from reading the RAPIDS docs is that manually reducing the `batch_size` in this call has a chance of fixing my OOM issues. Do we see a possibility of making this batch size user-configurable?

stevetracvc:
@HeadCase what version of cuML are you using? I think version 23.08 uses a batch size of 4096 which likely will fix your problem. You could also take a look at my first commit, which does include a batch_size parameter, if 4096 is too big for you

stevetracvc:
@MaartenGr sorry this dropped off my radar. I get it causing future problems, but rather than removing it I can add in version tests. I think it's 23.04 and 23.06 that need the batch_size patch

eg
`if cuml.__version__.startswith("23.04"):`

MaartenGr:
Seeing as the newest version of cuML resolves the issue. Would it not be straightforward to just mention that users will need to use the latest version? Especially since it was fixed in their latest release.

stevetracvc:
For me, it's a lot easier to update the BERTopic version in my environment than it is to update cuML. I've had major dependency conflicts when trying to install cuML with some of my other packages.

My point being, other people might not be able to update cuML in their environments just to get a feature that technically does work. But I get that you don't want the code cluttered. 

Your call. We could just leave the commit as is, and not merge it, and people can patch their copy of BERTopic if needed...

MaartenGr:
Hmmm, in that case, it might indeed be better to check for those specific versions and only implement the fix for these versions. I think that would be more stable and indeed still allows users to keep using the current version. As long as it does not affect newer versions, that should suffice.

Issue #1263, opened 2023-05-15
Issue with Azure OpenAI API usages with BERTopic


amitguptadumka:
I have an issue working with custom GPT model with Bertopic.
So i had used  default GPT integrated model for topic clustering as per documentation and it worked like charm(code below):

openai.api_key ='sk-xxxxxxxxxxxx'
representation_model= OpenAI(model="gpt-3.5-turbo", delay_in_seconds=15, chat=True)
topic_model = BERTopic(representation_model=representation_model)
topics5, probs5 = topic_model.fit_transform(docs)
This was via my personal API key i had experimented on public data.

Now  i am trying to use Bertopic with Azure integrated OPENAI API key which i am trying for enterprise data. 
There are custom OpenAI GPT model deployed by enterprise and when i use it with a chat completion model its giving proper response. But when i am using this API key and custom model with the initial code(representation model) its throwing me error , i am not sure how to convert this chat completion code and use it to create topic cluster. i tried approaches with documentation but no luck.
The error im getting with this is: InvalidRequestError: Must provide an 'engine' or 'deployment_id' parameter to create a <class 'openai.api_resources.chat_completion.ChatCompletion'>
Basically the custom model is nothing but OpenAI GPT 3.5 turbo model.

amitguptadumka:
Thanks. But as per this thread when i include custom model with 
from bertopic.representation import OpenAI
representation_model= OpenAI(engine="my-custom-embedding_model", delay_in_seconds=15, chat=True)

it does not accept "engine" as parameter.
TypeError: __init__() got an unexpected keyword argument 'engine'

MaartenGr:
You can use the `generator_kwargs` in the `OpenAI` module to specify additional kwargs for the ` .create` function:

```python
from bertopic.representation import OpenAI
representation_model= OpenAI(generator_kwargs={"engine":"my-custom-embedding_model")
```

amitguptadumka:
Thanks @MaartenGr . This worked like charm.

ninavdPipple:
I don't know if others encountered the same problem, but when I pass a 'model' in the generator_kwargs, I get an error that a certain arugment is passed twice into the function. If I look at the code, this actually makes sense as first it is unpacked in self.model and eventually both self.model and generator_kwargs are passed into the completions. I fixed this by changing 

        if self.generator_kwargs.get("model"):
            self.model = generator_kwargs.get("model")

to 

        if self.generator_kwargs.get("model"):
[...]

PR #1438, opened 2023-07-28
Update _openai.py

kulzen24:
Updated extract_topics function to take in argument "engine" instead of "model." This was erroring out for me as detailed in my latest issue, but was fixed on my local machine when I made this change.


MaartenGr:
I believe this should be solved with #1263 since this seems to be Azure-specific and might not work for general users. I would prefer specifically adding "engine" as mentioned in the issue above.

PR #1451, opened 2023-08-04
add support for claude, cohere and llama2

krrishdholakia:
Hi @MaartenGr, 

Saw you expanded support to OpenAI - and wanted to see how to add support for other LLM providers. I'm working on litellm (simple library to standardize LLM API Calls - https://github.com/BerriAI/litellm) and was wondering if we could be helpful.

Added support for Claude, Cohere, and Llama2 (via Replicate) by creating a new `_litellm.py` file. The code is pretty similar to the OpenAI class - as litellm follows the same pattern as the openai-python sdk.

Would love to know if this helps.

Happy to make any additional changes as required 😊

### DIFF
--- a/bertopic/representation/_litellm.py
+++ b/bertopic/representation/_litellm.py
@@ -1,0 +1,232 @@
+import time
+import litellm 
+from litellm import completion
+import pandas as pd
+from scipy.sparse import csr_matrix
+from typing import Mapping, List, Tuple, Any
+from bertopic.representation._base import BaseRepresentation
+from bertopic.representation._utils import retry_with_exponential_backoff
+
+
+DEFAULT_PROMPT = """
+This is a list of texts where each collection of texts describe a topic. After each collection of texts, the name of the topic they represent is mentioned as a short-highly-descriptive title
+---
+Topic:
+Sample texts from this topic:
+- Traditional diets in most cultures were primarily plant-based with a little meat on top, but with the rise of industrial style meat production and factory farming, meat has become a staple food.
+- Meat, but especially beef, is the word food in terms of emissions.
+- Eating meat doesn't make you a bad person, not eating meat doesn't make you a good one.
+
+Keywords: meat beef eat eating emissions steak food health processed chicken
+Topic name: Environmental impacts of eating meat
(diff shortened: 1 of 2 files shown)


MaartenGr:
Oops, I think you missed the [Contributing Guideliness](https://github.com/MaartenGr/BERTopic/blob/master/.github/CONTRIBUTING.md)! I think the idea of one package for many APIs is definitely welcome. Also, I see you are on a mission to add the functionality of `litellm` to many repos of which some have already merged, congrats!

Having said that, there are a few things that make me wonder. With the Langchain integration, which supports quite a number of LLMs, what is then the added value of `litellm`? Also, adding this as a main dependency would result in quite a number of additional dependencies to take into account whilst other LLM-based packages are not used as main dependencies, such as Langchain and OpenAI.

krrishdholakia:
hey @MaartenGr thanks for the feedback 

here's what i understand so far: 
- with `litellm` because we import all the llm packages at the top (openai, anthropic, replicate, etc.) this causes an issue re: dependencies, where you don't want to download packages you don't need 

Would a better experience here be to download packages as required (e.g. if the user is passing in an openai call, then download the package at that moment)? 

C

MaartenGr:
> Would a better experience here be to download packages as required (e.g. if the user is passing in an openai call, then download the package at that moment)?

As is done currently, which you can find in the [documentation](https://maartengr.github.io/BERTopic/getting_started/representation/llm.html#langchain), installation of the necessary packages should be done outside of BERTopic. Otherwise, it becomes too difficult to track all optional dependencies. The nice thing about this is that it easily allows for adding specific frameworks. 

> langchain is good because i can work across models, but i've found it bloated - adds a ton of dependencies that i don't need to my repo (e.g. SQLAlchemy). So wanted to figure out a way to do this in a simpler manner - which is why we built lite.

Ah right, that makes sense! Here, I am not sure how much it adds compared to what is already implemented. I am always a bit hesitant to add external packages that might not be necessary as it adds another layer of complexity with respect to dependency management. Having said that, if there are many users that would like to see this implemented, I am all for it!

PR #1466, opened 2023-08-10
Fix IndexError when all outliers are removed by reduce_outliers()

### LINKED ISSUE #1455: IndexError on `visualize_heatmap()` when all outliers are removed by `reduce_outliers()` and `update_topics()`
I encountered an IndexError with the visualization of heatmap. Here's the sequence of steps I took and the resulting error.
I instantiated a BERTopic model and ran the fit_transform() method as follows:
```
topic_model = BERTopic(
    embedding_model=embedding_model,
    vectorizer_model=vectorizer,
    calculate_probabilities=True,
    verbose=True,
    nr_topics=20,
)
topics, probs = topic_model.fit_transform(text_data, embeddings=embeddings)
```

Subsequently, I executed the following code:

```
new_topics = topic_model.reduce_outliers(text_data, topics, strategy="embeddings", embeddings=embeddings)
topic_model.update_topics(text_data, topics=new_topics, top_n_words=10)
```

After that, when I used `topic_model.visualize_heatmap()`, I encountered this IndexError:
```
File "<my directory>\bertopic\plotting\_heatmap.py", line 94, in visualize_heatmap
    embeddings = embeddings[indices]
IndexError: index 18 is out of bounds for axis 0 with size 18
```

I suspect the problem originates from `topic_model._outliers`. Before reducing outliers, the index was off by one compared to the topic number, as shown below (where the first column represents the index and the second column represents the topic number):
```
0   -1
1    0
2    1
3    2
... ...
19   18
```
After outlier reduction, it appears that all outliers were removed, resulting in the same index and topic number like below:
```
0   0
1   1
2   2
3   3
... ...
18   18
```

However, `topic_model._outliers` returned 1 for both patterns. In the `visualize_heatmap()` method of `_heatmap.py`, the embeddings array is sliced using `topic_model._outliers` like `embeddings = np.array(topic_model.topic_embeddings_)[topic_model._outliers:]`

Both patterns produce an `indices` array like `[ 0 1 2 3 ... 17 18]`. The shape of `embeddings` varies depending on the slice. With outliers, `embeddings.shape` is `(19, 768)`, but without outliers, it's `(18, 768)` because of slicing. This discrepancy causes the IndexError in embeddings = embeddings[indices].

I believe `topic_model._outliers` should be updated to 0 when all outliers are removed. A potential fix could involve adding the following line to the `update_topics()` method in `_bertopic.py`:
```
self._outliers = 1 if -1 in set(topics) else 0
```
I've tested this fix for my specific error and it seemed to resolve the issue. However, I haven't tested other parts of the library, so I'm unsure this is right solution. If it's appropriate, I'm willing to submit a PR.

Aratako:
Fix #1455
This fix addresses the IndexError occurring in update_topics() when all outliers are eliminated by reduce_outliers().

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -1380,6 +1380,8 @@ class BERTopic:
                           "manually assigning topics is the last step in the pipeline."
                           "Note that topic embeddings will also be created through weighted"
                           "c-TF-IDF embeddings instead of centroid embeddings.")
+
+        self._outliers = 1 if -1 in set(topics) else 0
         # Extract words
         documents = pd.DataFrame({"Document": docs, "Topic": topics, "ID": range(len(docs)), "Image": images})
         documents_per_topic = documents.groupby(['Topic'], as_index=False).agg({'Document': ' '.join})


MaartenGr:
It took a while to check but I believe this should not negatively affect topic assignment throughout the attributes I test. Thanks for the PR!

Issue #1527, opened 2023-09-13
Question/request about representative document truncation 


zilch42:
Hi Maarten,

I'm curious about the truncation at 255 characters of representative documents in both the `OpenAI` and `TextGeneration` representation models (at least). 

https://github.com/MaartenGr/BERTopic/blob/58d90bc65e95a11718e63a0834809d156dcf431d/bertopic/representation/_textgeneration.py#L143C1-L147C67

I imagine it's helpful if you are paying for an API, but if using a local text generation model e.g. as in your Llama2 google colab example, is there any other reason that the representative documents have to be cut down so hard? 

Could we have some control over the truncation e.g. add a `truncate: int = 255` parameter to the representation models and then:

```
for doc in docs:
    doc = doc[:truncate] if truncate else doc
    to_replace += f"- {doc}\n"
```

MaartenGr:
Yeah, I introduced that when LLMs had relatively small context sizes and as the trained documents are often relatively big (compared to context sizes) this was an easy limit to just use. Llama 2 models also have context sizes that limit the amount of text that you can give it. So you have to be really careful to not introduce too many tokens for just the documents. 

Having said that, opening up a parameter would be no problem here. I would actually prefer to have a truncate parameter that relates to the underlying tokenizer but unfortunately that is not easily possible without introducing many dependencies. Do you think it would suffice to use a character-based truncate parameter?

zilch42:
Yeah I he was thinking about that too. Is it generally true that when using `TextGeneration`, the user should have the tokenizer for the text generation model they want to use? In that case, the dependency is on the user not on BERTopic. 

Does `tiktoken` have a lot of dependencies? That covers `OpenAI` but I'm not sure what cohere uses. 

If not characters, would words be better because it's a bit easier for people to estimate tokens from words than it is from characters? E.g. for OpenAI 100 tokens ~ 75 words. It might also be useful too add to the documentation how many tokens the standard prompts e.g. for `OpenAI` amount to. Then you could have a default of maybe 50 words rather than 255 characters for shorter context length generation models.

MaartenGr:
> Yeah I was thinking about that too. Is it generally true that when using TextGeneration, the user should have the tokenizer for the text generation model they want to use? In that case, the dependency is on the user not on BERTopic and TextGeneration just needs to ask for it.
> 
> Does tiktoken have a lot of dependencies? That covers OpenAI but I'm not sure what cohere uses.

[...]

PR #1539, opened 2023-09-21
Add params to truncate documents to length when using LLMs

MaartenGr:
Truncate documents to a specific length when using any LLM for fine-tuning topic representations (#1527). It introduces two parameters:

* `doc_length`
    * The maximum length of each document. If a document is longer, it will be truncated. If None, the entire document is passed.
* `tokenizer`
    * The tokenizer used to calculate to split the document into segments used to count the length of a document. 
        * If tokenizer is  `'char'`, then the document is split up into characters which are counted to adhere to `doc_length`
        * If tokenizer is `'whitespace'`, the the document is split up into words separated by whitespaces. These words are counted       and truncated depending on `doc_length`
        * If tokenizer is `'vectorizer'`, then the internal CountVectorizer is used to tokenize the document. These tokens are counted and trunctated depending on `doc_length`
        * If tokenizer is a callable, then that callable is used to tokenized the document. These tokens are counted and truncated depending on `doc_length`

This means that the definition of `doc_length` changes depending on what constitutes a token in the `tokenizer` parameter. If a token is a character, then `doc_length` refers to max length in characters. If a token is a word, then `doc_length` refers to the max length in words.

The example below essentially states that documents cannot be longer than 100 tokens. Anything more than that will be truncated. 

It uses tiktoken which can be installed as follows together with openai:

```bash
pip install tiktoken openai bertopic
```

Example:

```python
import openai
import tiktoken
from bertopic.representation import OpenAI
from bertopic import BERTopic

# Tokenizer
tokenizer= tiktoken.encoding_for_model("gpt-3.5-turbo")

# Create your representation model
openai.api_key = MY_API_KEY
representation_model = OpenAI(
    model="gpt-3.5-turbo", 
    delay_in_seconds=2, 
    chat=True,
    doc_length=100,
    tokenizer=tokenizer
)

# Use the representation model in BERTopic on top of the default pipeline
```

### DIFF
--- a/bertopic/representation/_cohere.py
+++ b/bertopic/representation/_cohere.py
@@ -1,8 +1,10 @@
 import time
 import pandas as pd
+from tqdm import tqdm
 from scipy.sparse import csr_matrix
-from typing import Mapping, List, Tuple
+from typing import Mapping, List, Tuple, Union, Callable
 from bertopic.representation._base import BaseRepresentation
+from bertopic.representation._utils import truncate_document
 
 
 DEFAULT_PROMPT = """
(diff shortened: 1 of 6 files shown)


zilch42:
Hi Maarten,

Great stuff! Thanks for implementing all of the different split options. I have tested the TextGeneration representation model with llama2 with all of the tokenizer options and it all appears to be working. 

The only thing I'm not sure about is the last tokenizer case:

```
       elif isinstance(tokenizer, str):
            truncated_document = f"{tokenizer}".join(document.split(tokenizer)[:doc_length])
```

It isn't documented as an option and there is a risk that if a user misspells one of the other string based options (e.g. `tokenizer = "victorizer"`) then they may get an unexpected result rather than it raising a value error. Is there a specific use case it is solving? If so, it should probably be documented and I wonder if there is any way the user could be more explicit about providing a custom string to split on (or more explicit about the other string based options...)? And if not, does it need to be allowed? 

I appreciate the `truncate_document` function too. It makes testing the truncation a lot easier 😄

MaartenGr:
Thanks for the feedback! I forgot to add documentation for that it seems. It was initially created for those who wanted to separate a token based on a specific separate instead of using whitespace. Having said that, I think I will just remove it. It might make more sense to give additional instructions with respect to using `decode` and `encode` options that will support that type of behavior.

zilch42:
Before merging this, see #1545 with another potential use case for the split on character sequence option

PR #1616, opened 2023-11-04
Support saving/loading models that use OnlineCountVectorizer

### LINKED ISSUE #1431: Problem with saving the model
Hi, I am using the partial_fit function to perform incremental learning with BERTopic. When I tried to save the BERTopic model using safetensors, I got the following error: KeyError: 'tokenizer'. The error was raised in bertopic/_save_utils.py when the function tries to recreate the countvectorizer delete the parameters in cv but they don't actually exist. 
I tried to save the model using the code: model.save('some_directory', serialization="safetensors", save_ctfidf=True),
and here is the error code I got: 
/python3.9/site-packages/bertopic/_save_utils.py in save_ctfidf_config(model, path)
    293     # Recreate CountVectorizer
    294     cv_params = model.vectorizer_model.get_params()
--> 295     del cv_params["tokenizer"], cv_params["preprocessor"], cv_params["dtype"]
    296     if not isinstance(cv_params["analyzer"], str):
    297         del cv_params["analyzer"]

KeyError: 'tokenizer'

I have run the function model.vectorizer_model.get_params() and it only contains 2 parameters: {'decay': 0.05, 'delete_min_df': None}.
Is there anything I've done wrong? Thank you!

27pchrisl:
This PR adds support for serializing the config of the OnlineCountVectorizer by explicitly stating all the constructor properties. Based on the test, it looks like no other changes were needed to support the loading of the vectorizer.

Fixes #1431

### DIFF
--- a/bertopic/vectorizers/_online_cv.py
+++ b/bertopic/vectorizers/_online_cv.py
@@ -71,10 +71,45 @@ class OnlineCountVectorizer(CountVectorizer):
     def __init__(self,
                  decay: float = None,
                  delete_min_df: float = None,
-                 **kwargs):
+                 input="content",
+                 encoding="utf-8",
+                 decode_error="strict",
+                 strip_accents=None,
+                 lowercase=True,
+                 preprocessor=None,
+                 tokenizer=None,
+                 stop_words=None,
+                 token_pattern=r"(?u)\b\w\w+\b",
+                 ngram_range=(1, 1),
+                 analyzer="word",
+                 max_df=1.0,
+                 min_df=1,
+                 max_features=None,
+                 vocabulary=None,
+                 binary=False,
+                 dtype=np.int64,
+                 ):
         self.decay = decay
         self.delete_min_df = delete_min_df
-        super(OnlineCountVectorizer, self).__init__(**kwargs)
+        super(OnlineCountVectorizer, self).__init__(
+            input=input,
+            encoding=encoding,
+            decode_error=decode_error,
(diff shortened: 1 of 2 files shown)


MaartenGr:
Thanks for the PR! I am not entirely sure if adding the parameters manually is the right way to go here. What if the parameters of `CountVectorizer` get updated in a new release of scikit-learn? That would break this functionality. Instead, might it not be possible to update the `save_ctfidf_config` function to account for the missing parameters by looking at the internal params of the function, either by adding a function to save them or update the `get_params` function?

27pchrisl:
Hi @MaartenGr, no it is certainly not ideal! This was previously mentioned in https://github.com/scikit-learn/scikit-learn/issues/8822 and discussed further in https://github.com/scikit-learn/scikit-learn/issues/13555 - it seems for now this is the way others handle this.

I *could* add a test case that compared the params of `CountVectorizer` to a list in the test case, so if the list were changed a test would fail?

MaartenGr:
> I could add a test case that compared the params of CountVectorizer to a list in the test case, so if the list were changed a test would fail?

Although that would help detect any issues in newer versions, it will still break the code requiring an immediate update. It is a nice solution if there is not alternative but what about the following I mentioned in the previous message:

> Instead, might it not be possible to update the save_ctfidf_config function to account for the missing parameters by looking at the internal params of the function, either by adding a function to save them or update the get_params function?

This could be implemented as shown in the below example: https://stackoverflow.com/a/56491394/10532563

PR #1618, opened 2023-11-05
Support generating representative docs when using online modelling

27pchrisl:
The first time you fit an online model and request representative docs this works fine as every doc you fitted with will be in a cluster somewhere.

The second time you fit with a second set of docs, causing new clusters to be created, it's possible that none of the docs you passed fit into any of the existing clusters. This causes the `_extract_representative_docs` method to fail. Which results in you not being able to obtain representative docs for your new clusters.

This PR resolves this, so that `_extract_representative_docs` will return entries for all the topics, but some may be empty, based on the set of docs you provided.

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -3492,7 +3492,13 @@ class BERTopic:
             # Calculate similarity
             nr_docs = nr_repr_docs if len(selected_docs) > nr_repr_docs else len(selected_docs)
             bow = self.vectorizer_model.transform(selected_docs)
-            ctfidf = self.ctfidf_model.transform(bow)
+            try:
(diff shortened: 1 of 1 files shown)


MaartenGr:
Thank you for the PR! You might have missed the [CONTRIBUTING.md](https://github.com/MaartenGr/BERTopic/blob/master/.github/CONTRIBUTING.md) file where you can an overview of the steps we generally take when opening up a new PR. The reason why I link that file is that it is not clear to me what the exact issue is that you are fixing. Could you first create a reproducible example such that I can also see what the exact error/issue is? That way, it is easier to for me what the problem is and what could be potential ways to fix it.

27pchrisl:
Hi @MaartenGr, I wrote this up in an issue.

Fixes #1620

PR #1637, opened 2023-11-18
Fix TopicMapper example usage docstring

chrisji:
A `hdbscan_model` can't be used to construct a `TopicMapper`, so I've updated it to a nondescript `topics` variable to at least mitigate confusion. 

Perhaps the original intention was to provide `hdbscan_model.labels_` as the example, which may be a more informative example.


MaartenGr:
Thanks, LGTM!

PR #1701, opened 2023-12-18
hierarchical_topics: remove unused variable cluster_df significantly …

### LINKED ISSUE #1698: hierarchical_topics: unused variable cluster_df significantly slowing down performance
The variable `cluster_df` seems not to be used anywhere in the function: https://github.com/MaartenGr/BERTopic/blob/master/bertopic/_bertopic.py#L1008

The `groupby` call in it is the slowest line in the function too.

Maybe better remove it

shadiakiki1986:
…slowing down performance (closes #1698)


MaartenGr:
Thanks, LGTM!

Issue #1696, opened 2023-12-14
topic_model.transform(docs)[0][i] is sometimes different from topic_model.transform(docs[i])[0][0]


jonaslandsgesell:
Hello

I read https://maartengr.github.io/BERTopic/api/bertopic.html#bertopic._bertopic.BERTopic.transform and understood from the documents parameter (described as "A single document or a list of documents to predict on") that I could submit a list of documents or a single document and still receive the same result when predicting with a fitted model.

I found that this is not true. Below you find a minimal working example
`
from bertopic import BERTopic
from sklearn.datasets import fetch_20newsgroups

docs = fetch_20newsgroups(subset='all')['data'][:200]
topic_model = BERTopic().fit(docs)
topics,_=topic_model.transform(docs)

import numpy as np
topics=np.array(topics)

#calling the model with a single document several times
import tqdm

topics_single = []
for doc in tqdm.tqdm(docs):
    topic, _ = topic_model.transform([doc])
    topics_single.append(topic[0])
topics_single = np.array(topics_single)

mask_identical = topics_single == topics
percentage_equal = 100 * np.sum(mask_identical) / len(mask_identical)
print(f"{percentage_equal=}%") #returns for example about 60%, but varying

#loop till finding a different entry
for i in range(len(docs)):
    print(
        i,
        topic_model.transform(docs[i])[0][0],
        topic_model.transform([docs[i]])[0][0],
        topics[i],
        topic_model.transform(docs)[0][i],
    )
    if topic_model.transform(docs[i])[0][0] != topic_model.transform(docs)[0][i]:
        print(f"Different outcome at iteration {i}")
        break

`

The repeated execution with the same documents seems fine:
`topics2,_=topic_model.transform(docs)
percentage_equal_executed_with_multiple_docs = 100*np.sum(np.array(topics2)==topics)/len(topics)
print(f"{percentage_equal_executed_with_multiple_docs=}%")` #this gives 100%

The python version is 3.10.12
The list of installed packages is

`
absl-py==1.0.0
accelerate==0.23.0
adagio==0.2.4
aiohttp==3.8.6
aiosignal==1.3.1
ansi2html==1.9.1
antlr4-python3-runtime==4.11.1
anyio==3.5.0
appdirs==1.4.4
arch==6.2.0
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
astor==0.8.1
asttokens==2.0.5
astunparse==1.6.3
async-timeout==4.0.3
[...]

PR #1721, opened 2024-01-03
Update _bertopic.py to fix question/ github issue #1696

jonaslandsgesell:
As discussed in https://github.com/MaartenGr/BERTopic/issues/1696, I provide an updated doc string to reflect that `topic_model.transform(docs)[0][i]` is sometimes different from `topic_model.transform(docs[i])[0][0]`

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -456,7 +456,10 @@ class BERTopic:
         """ After having fit a model, use transform to predict new instances
 
         Arguments:
-            documents: A single document or a list of documents to predict on
+            documents: A single document or a list of documents to predict on.                    
+                       Note that the behavior of the method might differ depending 
+                       on whether a single document or a list of documents is passed
+                       (especially when using the HDBSCAN algorithm)
             embeddings: Pre-trained document embeddings. These can be used
                         instead of the sentence-transformer model.
             images: A list of paths to the images to predict on or the images themselves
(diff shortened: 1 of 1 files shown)


MaartenGr:
Thanks for this PR! Could you rephrase the following a bit:

> (especially when using the HDBSCAN algorithm)

This makes it seems that this behavior is across many different algorithms when in reality this is HDBSCAN-specific behavior.

jonaslandsgesell:
Sure! Do you have a suggestion for a specific wording?

MaartenGr:
> Sure! Do you have a suggestion for a specific wording?
>
> I am currently lacking the fantasy for other ways to express the fact that HDBSCAN is responsible here while we could also have a pipeline without HDBSCAN (but another component which may or may not behave similarly)

You could do something like this: *"A single document or a list of documents to predict the topic(s) for. NOTE: When using 
HDBSCAN, the prediction might differ depending on whether a single document or a list of documents is passed 
since it leverages the data points of other documents"*. 

I think it's best to stay close to the original documentation and inner workings of HDBSCAN. I believe [this](https://hdbscan.readthedocs.io/en/latest/soft_clustering.html) and [this](https://hdbscan.readthedocs.io/en/latest/soft_clustering_explanation.html) resource are relevant from the top of my head.

Also, a small tip. ChatGPT works wonders for helping with these kinds of issues ;)

Issue #1720, opened 2024-01-02
flan-t5 for 


mjaved-nz:
Hi @MaartenGr,

I hope you are doing well. I am getting the following error when using the flan-t5 model for topic representation. Any solution for this? Thanks

```
from transformers import pipeline
from bertopic.representation import TextGeneration

prompt = "I have a topic described by the following keywords: [KEYWORDS]. Based on the previous keywords, what is this topic about?"

# Create your representation model
generator = pipeline('text2text-generation', model='google/flan-t5-base')
representation_model = TextGeneration(generator)

topic_model_t5 = BERTopic(representation_model=representation_model)
```

Error:

----> 2 topics, probs = topic_model_t5.fit_transform(docs)
      3 print(topic_model_t5.get_topic_info())

[/usr/local/lib/python3.10/dist-packages/bertopic/representation/_textgeneration.py](https://localhost:8080/#) in extract_topics(self, topic_model, documents, c_tf_idf, topics)
    145 
    146             # Prepare prompt
--> 147             truncated_docs = [truncate_document(topic_model, self.doc_length, self.tokenizer, doc) for doc in docs]
    148             prompt = self._create_prompt(truncated_docs, topic, topics)
    149             self.prompts_.append(prompt)

TypeError: 'NoneType' object is not iterable

MaartenGr:
It seems like it cannot iterate over the documents for whatever reason. Did you make sure that all documents are no-empty? That also means that very short documents that contain for instance `"\n"` should also be removed.

mjaved-nz:
I don't have any empty documents the minimum length of the documents is 21. The same set of documents works fine with other LLMs.

MaartenGr:
> I don't have any empty documents the minimum length of the documents is 21. 

How did you calculate the length of the document? Tokenization schemes of the underlying model might handle certain documents differently. 

> The same set of documents works fine with other LLMs.

Which other LLMs did you try? Did it work with `TextGeneration` or something else?

Also, on how many documents did you train your model? It might be that there are only a couple of documents per topic and that it might not properly return a document.

manveersadhal:
Hi @MaartenGr - thank you for creating this fantastic library.

I think the cause is that when the DEFAULT_PROMPT is used (which has no [DOCUMENTS]) or a user-supplied prompt does not contain "[DOCUMENTS]", the docs in repr_docs_mappings are all assigned a value of None. The error occurs when trying to iterate over None.

[...]

PR #1726, opened 2024-01-06
Fix bug in TextGeneration prompt creation when prompt template does not contain [DOCUMENTS]

manveersadhal:
Prevent attempted iteration over NoneType when prompt template does not contain [DOCUMENTS] in TextGeneration class.

When DEFAULT_PROMPT is used or the user-provided prompt does not contain "[DOCUMENTS]", all of the values in repr_docs_mappings are None. When defining truncated_docs, iteration is attempted over NoneType. This change prevents the attempted iteration and instead passes the None value to _create_prompt.

### DIFF
--- a/bertopic/representation/_textgeneration.py
+++ b/bertopic/representation/_textgeneration.py
@@ -144,7 +144,7 @@ class TextGeneration(BaseRepresentation):
         for topic, docs in tqdm(repr_docs_mappings.items(), disable=not topic_model.verbose):
 
             # Prepare prompt
-            truncated_docs = [truncate_document(topic_model, self.doc_length, self.tokenizer, doc) for doc in docs]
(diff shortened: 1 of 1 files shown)


manveersadhal:
This is to address issue #1720

MaartenGr:
Thanks for sharing this fix! Do you by chance have a reproducible example that allows me to test whether this test works for the issue that you are facing?

leoschet:
Any idea when this will be released?

MaartenGr:
@leoschet It is merged, so you can install BERTopic from the main branch. An official release might take a while since there are a couple PRs open that need to be merged/checked.

leoschet:
I've been doping that, just not always possible to pin main for the package version.

PR #1733, opened 2024-01-10
Fix _openai.py representation #1263

ninavandiermen:
fix: prevent 'model' argument being passed twice when using 'generator_kwargs'

'model' is deleted from 'generator_kwargs' after it is obtained from it.


MaartenGr:
Thanks for the PR, LGTM!

PR #1750, opened 2024-01-15
Add optional datamapplot visualization

lmcinnes:
A basic version of using DataMapPlot for visualization. Documentation is just the API as I wasn't sure quite what you might want or need in the "getting started" docs for such a thing. Let me know if you have any requests or suggestions.


bertopic/_bertopic.py
```diff
@@ -2339,6 +2339,95 @@ def visualize_documents(self,
...
+                                   reduced_embeddings: np.ndarray = None,
+                                   custom_labels: Union[bool, str] = False,
+                                   title: str = "Documents and Topics",
+                                   sub_title: Union[str, None] = None,
+                                   width: int = 1200,
+                                   height: int = 1200,
+                                   **datamap_kwds):
+        """ Visualize documents and their topics in 2D as a static plot for publication using
+        DataMapPlot. This works best if there are between 5 and 60 topics. It is therefore best
+        to use a sufficiently large `min_topic_size` or set `nr_topics` when building the model.
```
MaartenGr, line 2355: What do you think about adding a `topics` parameter that allows you to select a subset of topics to visualize? That way, users are not limited by needing to generate between 5 and 60 topics but can freely select whichever topics they want visualized. All other topics would then be shown as -1/outliers/unlabelled I guess.

bertopic/plotting/_datamap.py
```diff
@@ -0,0 +1,134 @@
...
+def visualize_document_datamap(topic_model,
+                               docs: List[str],
+                               embeddings: np.ndarray = None,
+                               reduced_embeddings: np.ndarray = None,
+                               custom_labels: Union[bool, str] = False,
+                               title: str = "Documents and Topics",
+                               sub_title: Union[str, None] = None,
+                               width: int = 1200,
+                               height: int = 1200,
+                               **datamap_kwds) -> Figure:
```
MaartenGr, line 21: This throws an error if Figure was not imported from matpotlib if it was not installed. Which is why the testing pipeline failed.

bertopic/_bertopic.py
```diff
@@ -2339,6 +2339,95 @@ def visualize_documents(self,
                                             width=width,
                                             height=height)
 
+    def visualize_document_datamap(self,
```
MaartenGr, line 2342: Thanks for adhering to my (somewhat) unconventional method of accessing the plotting capabilities!

bertopic/plotting/_datamap.py
```diff
@@ -0,0 +1,134 @@
...
+
+    topic_name_mapping = {topic_num: topic_name for topic_num, topic_name in zip(unique_topics, names)}
+    topic_name_mapping[-1] = "Unlabelled"
+    named_topic_per_doc = pd.Series(topic_per_doc).map(topic_name_mapping).values
+
+    figure, axes = datamapplot.create_plot(
+        embeddings_2d,
+        named_topic_per_doc,
+        figsize=(width/100, height/100),
+        dpi=100,
```
MaartenGr, line 128: Out of curiosity, why did you set this at a default of 100? Is the default `dpi` not sufficient for these plots?

MaartenGr (commented):
Thank you for not only for this PR but also your work on UMAP, HDBSCAN, and This Not That! Truly wonderful and an amazing package for sharing annotated 2-dimensional embeddings. 

With respect to the additional documentation in the examples, I think this deserves a spot but I am not sure where. Perhaps the quickstart here: https://maartengr.github.io/BERTopic/getting_started/quickstart/quickstart.html

But I can also add that later. 

Lastly, at some point this should be integrated when saving a BERTopic model to the Hugging Face Hub. Having these images added to a readme would be great but it might take some effort to add that. I can check out what is necessary after this PR.

bertopic/_bertopic.py
lmcinnes: That's a reasonable option, and I agree that it makes sense. Let me see if I can make that work.

bertopic/plotting/_datamap.py
lmcinnes: That makes sense; my tests ran locally, but of course matplotlib was installed. I think I can wrap the import in a try accept and have a dummy class for other case.

bertopic/plotting/_datamap.py
lmcinnes: 100 is the default DPI for matplotlib. To keep with "width x height" in pixels as input params (following the other plotting methods) it helps if the DPI stays fixed, so I set it to the default.

bertopic/plotting/_datamap.py
lmcinnes: This should work now. See [f8104c2](https://github.com/MaartenGr/BERTopic/pull/1750/commits/f8104c202ef57cddfac6f1da287521f316ee9a59).

MaartenGr:
Amazing, thanks for the hard work, both on the DataMapPlot package as well as this PR. It is merged! In the upcoming weeks, I plan to add it to the official documentation (https://maartengr.github.io/BERTopic/index.html) but I'll have to check the best location for it so it does not get buried in the docs.

Issue #1755, opened 2024-01-16
representation_model: 'NoneType' object is not iterable


muehlhausen:
Hey there!

First of all: thank you for developing BERTopic, it's neat! However, I am encountering an issue with `representation_model`, when trying to rename my cluster representations. Everything works fine as long as I am using just an embedding_model. However, as soon as I start using a representation_model I get the same error consistently.

Here is some sample code, inspired by this [documentation](https://maartengr.github.io/BERTopic/getting_started/representation/llm.html#transformers). 

```
# Import the necessary libraries
from bertopic import BERTopic
import pandas as pd
from transformers import pipeline
from bertopic.representation import TextGeneration

# prompt = f"I have a topic described by the following keywords: [KEYWORDS]. Based on the previous keywords, what is this topic about?"

# Create your representation model
generator = pipeline('text2text-generation', model='google/flan-t5-base')
representation_model = TextGeneration(generator)

# 4. Get some sample data
data = pd.read_excel(testdata.xlsx')

# 5. Initialize BERTopic with the representation model
topic_model = BERTopic(
    embedding_model= 'paraphrase-multilingual-mpnet-base-v2',
    representation_model = representation_model # if commented, code works
)

# 6. Fit BERTopic to the sample texts
topics, _ = topic_model.fit_transform(data['text'])

# 6. Get the topic information
topic_info = topic_model.get_topic_info()

# 7. Print the topic information
print(topic_info)
```

The error I get is:

```
TypeError                                 Traceback (most recent call last)
Cell In[3], line 26
     20 topic_model = BERTopic(
     21     embedding_model= 'paraphrase-multilingual-mpnet-base-v2',
     22     representation_model = representation_model
     23 )
     25 # 6. Fit BERTopic to the sample texts
---> 26 topics, _ = topic_model.fit_transform(data['Absatz'])
     28 # 6. Get the topic information
     29 topic_info = topic_model.get_topic_info()

File ~/Code/NDR/.venv/lib/python3.11/site-packages/bertopic/_bertopic.py:433, in BERTopic.fit_transform(self, documents, embeddings, images, y)
    430     self._save_representative_docs(custom_documents)
    431 else:
    432     # Extract topics by calculating c-TF-IDF
--> 433     self._extract_topics(documents, embeddings=embeddings, verbose=self.verbose)
    435     # Reduce topics
    436     if self.nr_topics:

[...]

PR #1756, opened 2024-01-16
fix: text generation without repo docs

leoschet:
Related to: https://github.com/MaartenGr/BERTopic/issues/1755

### DIFF
--- a/bertopic/representation/_textgeneration.py
+++ b/bertopic/representation/_textgeneration.py
@@ -138,7 +138,7 @@ class TextGeneration(BaseRepresentation):
                 self.diversity
             )
         else:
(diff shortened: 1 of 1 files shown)


leoschet:
While at it, I would also propose considering a change to the respective if condition. we should only check if `[DOCUMENTS]` is in `self.prompt`, this way the code is robust for future changes to the default prompt

leoschet:
Well, I missed the other [PR opened for the same issue](https://github.com/MaartenGr/BERTopic/pull/1726). @MaartenGr feel free to proceed as you see fit, of course.

MaartenGr:
Thanks for the PR, I'll close this in favor of the other that was already opened.

PR #1774, opened 2024-01-29
Remove spurious warning about missing embedding model

sliedes:
It's possible to specify an embedding model when using BERTopic.load(). Even when that is done, a warning was issued about not defining an embedding model. Silence that warning in this case.


MaartenGr:
Thanks, LGTM!

PR #1803, opened 2024-02-10
Fix type hint in ClassTfidfTransformer constructor

snape:
I see only examples of integers being used for `seed_multiplier` in the documentation and for the default, but see no reason why a floating point number could not be used. In any case, `bool` is inconsistent and presumably a typographic error. 

(Please excuse the lack of creating an issue in advance of this very minor fix.)


MaartenGr:
Thanks! LGTM 😄

Issue #1798, opened 2024-02-08
topic representations after `merge_models` is consistently wrong on unseen topics


Neehier:
Not sure what's going on, wondering whether this is a bug or just a mistake in my code.

Basically after having merged two models using `merge_models`, the predicted **topic** using `transform` is consistently incorrect only if an unseen topic (new after merging) is predicted. The interesting thing is that all items that are predicted to this new unseen topic, do actually relate to each other and in fact belong to a different topic. Sometimes this topic is one of the new topics, and other times it is one of the older topics.

This makes me think something is going wrong somewhere with the sorting of topics. I've tried taking a look using `get_topic_info()` but the representations linked to the topics that are wrongfully predicted, do indeed not match the topics in the case of new topics.

```py
# merge models
tuned_model = BERTopic.load(PATH-NEW)
tuned_model_base = BERTopic.load(PATH-BASE)

merged_model = BERTopic.merge_models([tuned_model_base, tuned_model])

# predict topics
res = pd.read_csv(PATH-RES)
chunks = [res[i:i+50000] for i in range(0, res.shape[0], 50000)]

preds = []
probs = []

for df in chunks:
    docs = df.text.to_list()
    pred = merged_model.transform(docs)
    preds.extend(pred[0])
    probs.extend(pred[1])

# assign topics to items
res['Topic'] = preds
res['Prob'] = [max(probs[i]) for i in range(len(probs))]

inf = merged_model.get_topic_info()
topic_sizes = pd.DataFrame(list(merged_model.topic_sizes_.items()), columns=['Topic', 'Size'])

df = res.merge(inf[['Topic', 'Representation']], on='Topic', how='left')
df = df.merge(topic_sizes, on='Topic', how='left')
```

I'm afraid I cannot share output, though if I'd try to filter `df` for any topic that was unseen previous to the merging of models, then all the texts would be correctly clustered together, but assigned the wrong topic.

The models had been saved using the `pytorch` serialization method, and have been configured as such:

```py
umap_model = UMAP(n_components=x, n_neighbors=x, min_dist=x)
hdbscan_model = HDBSCAN(min_cluster_size=x, min_samples=x, prediction_data=True)
vectorizer_model = CountVectorizer(min_df=x, stop_words="english")
ctfidf_model = ClassTfidfTransformer(bm25_weighting=False, reduce_frequent_words=False)
representation_model = KeyBERTInspired()
```

Neehier:
[...]

PR #1804, opened 2024-02-12
Fix #1797 and #1798

### LINKED ISSUE #1797: Using `transform` with zeroshot
Hi Maarten. Thanks for all the work on this brilliant library.

I think I’ve identified a bug in the zeroshot feature. If I follow [the tutorial](https://maartengr.github.io/BERTopic/getting_started/zeroshot/zeroshot.html) and compare the results to a `transform`:

```
from pandas import DataFrame

from bertopic import BERTopic

# We select a subsample of 5000 abstracts from ArXiv
docs = dataset["abstract"][:5_000]

zeroshot_topic_list = ["Clustering", "Topic Modeling", "Large Language Models"]

# We fit our model using the zero-shot topics
# for clustering instead.
topic_model = BERTopic(
    min_topic_size=15,
    zeroshot_min_similarity=.85,
    representation_model=KeyBERTInspired()
)
topics, probabilities = topic_model.fit_transform(docs)

# Calculate topics and probabilities using the transform method
new_topics, new_probabilities = topic_model.transform(docs)

result = DataFrame(data={"documents": docs, "topics": topics, "probabilities": probabilities, "new_topics": new_topics, "new_probabilities": new_probabilities})
```

When I look at `result.head(10)` I see different results:

Screenshot 2024-02-08 at 10 44 49

From what I understand the results shouldn’t be identical, because `fit_transform` uses the topics defined by the original models (UMAP, HDBSCAN, etc.), while `transform` uses cosine similarity with topic embeddings. However, a difference of 2 seems weirdly common. If I look at the average difference between the two predictions `(result["topics"] - result["new_topics"]).value_counts()`, most of the time the difference is -2:

```
-2     3114
 0      353
-9       77
-25      58
 2       55
       ... 
 31       1
-50       1
 26       1
 19       1
 52       1
Name: count, Length: 102, dtype: int64
```

I think that this is because `topic_model.topic_embeddings_` are incorrect: they seem to be off by two positions. Perhaps this is related to [this issue](https://github.com/MaartenGr/BERTopic/issues/1771).

As a workaround I’m reidentifying the outlier topic and recalculating topic embeddings:

```
from bertopic._bertopic import TopicMapper

embeddings = topic_model.embedding_model.embed(docs)

topic_info = topic_model.get_topic_info()
outlier_label = topic_info.loc[topic_info["Name"].str.startswith("-1"), "Topic"].values[0]

# Restructure topic labels, sizes, representations and representative documents
for attribute_name in ["topic_labels_", "topic_sizes_", "topic_representations_", "representative_docs_"]:
    attribute = getattr(topic_model, attribute_name)
    attribute[-1] = attribute.pop(outlier_label)
    attribute = {key if key < outlier_label else key - 1: value for key, value in attribute.items()}
    attribute = dict(sorted(attribute.items()))
    setattr(topic_model, attribute_name, attribute)

topics = [topic if topic < outlier_label else -1 if topic == outlier_label else topic - 1 for topic in topics]

# Recreate topic embeddings
topic_model.topic_embeddings_ = None
topic_model.topic_mapper_ = TopicMapper(list(topic_model.topic_labels_.keys()))  # The topic mapper is used to map topics between merged models, we override it here to avoid errors
topic_model._create_topic_vectors(documents=DataFrame(data={"sentences": docs, "Topic": topics, "ID": range(len(docs))}), embeddings=embeddings)

topic_model._outliers = 1  # We have outliers in the model
```

The results look sensible:

```
new_topics, new_probabilities = topic_model.transform(docs)
result = DataFrame(data={"documents": docs, "topics": topics, "probabilities": probabilities, "new_topics": new_topics, "new_probabilities": new_probabilities})
```

`result.head(10)`:

Screenshot 2024-02-08 at 10 41 44

`(result["topics"] - result["new_topics"]).value_counts()`:

```
 0     3470
-10      75
-24      59
-11      57
-26      55
       ... 
 47       1
 17       1
 14       1
 27       1
 54       1
Name: count, Length: 107, dtype: int64
```

MaartenGr:
A fix for #1797 and #1798. As mentioned in these links, there were 2 issues:
* Embeddings not ordered correctly when using `.merge_models`
* Outlier topic not in the 0th position when using zero-shot topic modeling causing prediction issues (amongst others)

The embeddings should now be ordered correctly and the outlier topic is not at the 0th position when using zero-shot topic modeling. This should make it possible to use zero-shot topic modeling with outlier reduction and other techniques.

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -3237,7 +3237,7 @@ class BERTopic:
                     merged_topics["topic_aspects"][str(new_topic_val)] = selected_topics["topic_aspects"][str(new_topic)]
 
                 # Add new embeddings
-                new_tensors = tensors[new_topic - selected_topics["_outliers"]]
+                new_tensors = tensors[new_topic + selected_topics["_outliers"]]
                 merged_tensors = np.vstack([merged_tensors, new_tensors])
 
             # Topic Mapper
@@ -3663,6 +3663,32 @@ class BERTopic:
         self.__dict__.clear()
         self.__dict__.update(merged_model.__dict__)
         logger.info("Zeroshot Step 3 - Completed \u2713")
+
+        # Move -1 topic back to position 0 if it exists
+        if self._outliers:
+            nr_zeroshot_topics = len(set(y))
+
+            # Re-map the topics such that the -1 topic is at position 0
+            new_mappings = {}
+            for topic in self.topics_:
+                if topic < nr_zeroshot_topics:
+                    new_mappings[topic] = topic
+                elif topic == nr_zeroshot_topics:
+                    new_mappings[topic] = -1
+                else:
+                    new_mappings[topic] = topic - 1
+
+            # Re-map the topics including all representations (labels, sizes, embeddings, etc.)
+            self.topics_ = [new_mappings[topic] for topic in self.topics_]
+            self.topic_representations_ = {new_mappings[topic]: repr for topic, repr in self.topic_representations_.items()}
+            self.topic_labels_ = {new_mappings[topic]: label for topic, label in self.topic_labels_.items()}
+            self.topic_sizes_ = collections.Counter(self.topics_)
+            self.topic_embeddings_ = np.vstack([
+                self.topic_embeddings_[nr_zeroshot_topics],
+                self.topic_embeddings_[:nr_zeroshot_topics],
+                self.topic_embeddings_[nr_zeroshot_topics+1:]
+            ])
+
         return self.topics_
 
     def _guided_topic_modeling(self, embeddings: np.ndarray) -> Tuple[List[int], np.array]:
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -3660,12 +3660,13 @@ class BERTopic:
         merged_model.topics_ = df.Label.values
 
         # Update the class internally
+        has_outliers = bool(self._outliers)
         self.__dict__.clear()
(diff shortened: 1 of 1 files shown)


timdhu (approved):
I've tested the two examples in #1797 and #1798, both output the correct results following these changes.

bertopic/_bertopic.py
```diff
@@ -3663,6 +3663,32 @@ def _combine_zeroshot_topics(self,
         self.__dict__.clear()
         self.__dict__.update(merged_model.__dict__)
```
timdhu, line 3664: I spoke too soon! `merged_model.__dict__` replaces `self.outliers` with 0, so the outliers are never repositioned a couple of lines down.

timdhu: Maybe it's just best to set `outliers = self._outliers` just before merging on line 3627, then using this on line 3668 instead?

Will also have to set `self._outliers = 1` inside the if statement

timdhu: I tested this out and it looks good to me

outliers

MaartenGr: Awesome, thanks for checking this out! I just pushed a commit which should finally resolve this issue. If you could do a final check, I can merge this PR.

MaartenGr:
@timdhu Thanks for the help fixing this issue. Your dedication to finding the underlying cause as well as testing/suggesting the needed changes are greatly appreciated!

PR #1850, opened 2024-02-29
remove commas from pip install reference in readme

luisoala:
with the commas it will throw an error when installing


MaartenGr:
Thanks for the PR. LGTM!

PR #1874, opened 2024-03-19
Fix CUML HDBSCAN predictions by using correct method.

Bougeant:
This fix is intended to fix the bug found here: https://github.com/MaartenGr/BERTopic/discussions/1463

### DIFF
--- a/bertopic/cluster/_utils.py
+++ b/bertopic/cluster/_utils.py
@@ -51,8 +51,8 @@ def hdbscan_delegator(model, func: str, embeddings: np.ndarray = None):
 
         str_type_model = str(type(model)).lower()
         if "cuml" in str_type_model and "hdbscan" in str_type_model:
-            from cuml.cluster.hdbscan.prediction import approximate_predict
-            probabilities = approximate_predict(model, embeddings)
+            from cuml.cluster.hdbscan.prediction import membership_vector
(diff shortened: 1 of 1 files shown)


MaartenGr:
Thanks for the PR! Just to be sure I understand correctly, why is this specific change a solution to the problem? Also, which version of cuML are you using? I believe there were some changes to recent versions that might have fixed this.

Bougeant:
So I think when `func == "membership_vector"`, we should calculate probabilities using `membership_vector(model, embeddings)`, not using `approximate_predict(model, embeddings)` which merely calculates the most relevant cluster and the associated probability. 

I believe that this was just a typo.

MaartenGr:
@Bougeant I believe this relates to https://github.com/MaartenGr/BERTopic/pull/1324 where `approximate_predict` was used for earlier versions of cuML compared to newer versions which do use `membership_vector`. So this was not a typo but merely the functionality that was in v23.03 and lower I believe. If I'm not mistaken, `membership_vector` was introduced in v23.04. 

Perhaps that related PR should be updated and merged instead?

Bougeant:
Oh right, makes sense. In any case, the current implementation does not work since we expect a 2D array but are getting a 2-tuple. #1324 adds some unit tests as well, so that's probably the way to go.

Issue #1884, opened 2024-03-22
Multimodal imaging - documentation needs updating for Pandas change


mepearson:
Note: Will submit pull request with updated documents.

Error: 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[12], line 1
----> 1 df.drop("Representative_Docs", 1).drop("Name", 1)

TypeError: DataFrame.drop() takes from 1 to 2 positional arguments but 3 were given
```

Assumed reason: changes to pandas require use of keyword.

Solution: Change code to
df = topic_model.get_topic_info().drop(["Representative_Docs","Name"], axis=1)

PR #1885, opened 2024-03-22
Update multimodal.md to add axis keyword [Issue #1884]

mepearson:
Update Multimodal documentation to add an axis keyword for the drop function. This addresses issue #1884

Existing Code: 
```
# Extract dataframe
df = topic_model.get_topic_info().drop("Representative_Docs", 1).drop("Name", 1)
```

Suggested Code:
`# Extract dataframe
df = topic_model.get_topic_info().drop(["Representative_Docs","Name"], axis=1)`

This corrects the following error caused by the need for keyword arguments in newer versions of pandas. Columns to drop were gathered into a single list just for simplicity.
`---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[12], line 1
----> 1 df.drop("Representative_Docs", 1).drop("Name", 1)

TypeError: DataFrame.drop() takes from 1 to 2 positional arguments but 3 were given`

### DIFF
--- a/docs/getting_started/multimodal/multimodal.md
+++ b/docs/getting_started/multimodal/multimodal.md
@@ -64,7 +64,7 @@ def image_formatter(im):
     return f'<img src="data:image/jpeg;base64,{image_base64(im)}">'
 
 # Extract dataframe
-df = topic_model.get_topic_info().drop("Representative_Docs", 1).drop("Name", 1)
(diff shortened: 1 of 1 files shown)


MaartenGr:
Thanks for the PR! Do you know why this change is necessary? The reason why I ask is that this might relate to specific versions of pandas as this piece of code did work before. Did you test this?

PR #1894, opened 2024-03-28
Hierachical topics calculated with topic embeddings

azikoss:
Added an option to use topic embeddings to calculate distances between clusters for hierarchical topics. The topic embeddings should provide a better representation of the term frequencies calculated based on c-TF-IDF.


MaartenGr:
Thank you for this PR! I might be mistaken but I think you might have missed the [CONTRIBUTING](https://github.com/MaartenGr/BERTopic/blob/master/.github/CONTRIBUTING.md) file since I'm not seeing an attached issue. 

Having said that, we can start this feature request here for now. 

I definitely agree that having the option to switch between c-TF-IDF and topic embeddings would be great to have, not only in this specific example but throughout BERTopic. Keeping that in mind, I actually think it would be best that whenever such a parameter (`use_ctfidf`) is introduced, it is also implemented wherever relevant. For instance, having this feature in `.topics_over_time`, `.topics_per_class`, `.visualize_hierarchy`, and even `._reduce_to_n_topics`/`._auto_reduce_topics`. 

So I think this PR might be a bit bigger than the single function. I believe that if this parameter is in this single function, users will expect it to be used in others as well.

azikoss:
Hi @MaartenGr, my apologies for not reading the contributing file.

I am happy to make changes in the other methods as well. The list of the methods above is complete or they are only examples?

MaartenGr:
Thanks, that would be great! A few things to keep in mind:

* The `._reduce_to_n_topics`/`._auto_reduce_topics` might be a bit tricky as it is a private function, so the `use_c_tf_idf` param does not work there and it is not a parameter that fits with `__init__`
* Although `use_c_tf_idf` is the most logical, I think it's a nicer experience to type `use_ctfidf` instead
* I believe, but you will have to check, that the following also can choose between them:
  * `bertopic.plotting.visualize_heatmap`
  * `bertopic.plotting.visualize_hierarchy`
  * `bertopic.plotting.visualize_topics`
  * `bertopic.BERTopic.topics_over_time`
  * `bertopic.BERTopic.topics_per_class`
* Note that even if you choose ctfidf, it might not always be available, and the same goes for the topic embeddings. So it should revert to whatever is available
* The `topics_over_time` and `topics_per_class` are quite tricky, so I think it's okay to skip these for now
* Some testing might be required to check if everything works as intended since there might be some unexpected issues here and there (for instance, I've had issues in the past with the distance function in `hierarchical_topics` using topic embeddings

Hopefully, it is not too much. But if it is, please let me know! I'll see if I can help out.

azikoss:
Thanks for the guidelines! I will look into it within the next two weeks.

azikoss:
@MaartenGr, please have a look. 

I included the embedding selection to all mentioned methods apart from `topics_over_time(..)` and `topics_per_class(..)`.

I left the defaults for attribute `use_ctfidf` according to the current implementations. For instance `visualize_heatmap(..)` has `use_ctfidf = False` while `visualize_hierarchy(..)` use_ctfidf = True. This is the current behavior, but it would be nice to unify it - i.e. use semantic topic embeddings wherever possible. What do you think?

While the function `select_topic_represenation(..)` is unit tested, I did not add any end-to-end tests. Let me know what you think about it. 


`select_topic_represenation(..)` always returns an `np.ndarray` - it seems to be the general pattern throughout the code with some exceptions that I would like to point out. E.g., `plotting.visualize_hierarchy(..)` only converts `topic_embeddings_` into `ndarray` and not `c_tfidf` representation. In `plotting.visualize_topics(..), only c_tfidf are converted into `ndarray`.

MaartenGr:
@azikoss Thanks for the work on this! Hopefully, I will have some time this weekend or beginning of next week to look at this a bit more in-depth. 

> I left the defaults for attribute use_ctfidf according to the current implementations. For instance visualize_heatmap(..) has use_ctfidf = False while visualize_hierarchy(..) use_ctfidf = True. This is the current behavior, but it would be nice to unify it - i.e. use semantic topic embeddings wherever possible. What do you think?

I agree. Unification would definitely be best here. It will likely change the output so it might be nice to additionally test this. However, since there is the option to easily switch between type of topic representation embedding I'm not too worried about this. 

> While the function select_topic_represenation(..) is unit tested, I did not add any end-to-end tests. Let me know what you think about it.

I just started the tests, so let's see what happens with end-to-end tests using the defaults. In an ideal world, I would like to see more tests that also cover this end-to-end, but they are already quite large and slow. 

> select_topic_represenation(..) always returns an np.ndarray - it seems to be the general pattern throughout the code with some exceptions that I would like to point out. E.g., plotting.visualize_hierarchy(..) only converts topic_embeddings_ into ndarray and not c_tfidf representation. In plotting.visualize_topics(..), only c_tfidf are converted into ndarray`.

I would have to check this but I generally want to prevent casting sparse matrices into numpy arrays as that can increase the memory needed to hold that matrix significantly.

azikoss:
@MaartenGr thanks for your comments. 

I just ran the test on Python 3.9. They all run successfully. If I understand correctly, the support is also for 3.8?

Have a look at the conversion to np.array pls and let me know. It is about 50/50 where the c_tfidf matrix is converted to np.array and where it is not.

bertopic/_utils.py
```diff
@@ -147,3 +147,57 @@ def validate_distance_matrix(X, n_samples):
...
+
+    Returns:
+        The selected topic representation and a boolean indicating whether it is c-TF-IDF.
+    """
+
+    def to_ndarray(array: Union[np.ndarray, csr_matrix]) -> np.ndarray:
+        if isinstance(array, csr_matrix):
+            return array.toarray()
+        elif not isinstance(array, np.ndarray):
+            raise ValueError("The embeddings should be either a type of `numpy.ndarray` or `scipy.sparse.csr_matrix`")
```
MaartenGr, line 177: Why are you checking for this? The input is always either np.array or csr_matrix, so this should not happen in the first place right?

bertopic/_utils.py
```diff
@@ -147,3 +147,57 @@ def validate_distance_matrix(X, n_samples):
...
+            - If no topic representation was found
+            - If c-TF-IDF embeddings are not a numpy array or a scipy.sparse.csr_matrix
+
+    Returns:
+        The selected topic representation and a boolean indicating whether it is c-TF-IDF.
+    """
+
+    def to_ndarray(array: Union[np.ndarray, csr_matrix]) -> np.ndarray:
+        if isinstance(array, csr_matrix):
+            return array.toarray()
```
MaartenGr, line 175: I would advise not casting a csr_matrix at all unless the original implementation already does so. The reason for this is that I only cast to numpy unless absolutely necessary. Some functions/models can deal with sparse matrices but not all of them.

bertopic/_bertopic.py
```diff
@@ -928,12 +929,14 @@ def hierarchical_topics(self,
 
         Arguments:
             docs: The documents you used when calling either `fit` or `fit_transform`
+            use_ctfidf: Whether to calculate distances between topics based on c-TF-IDF embeddings. If False, semantic
+                        embeddings are used.
```
MaartenGr, line 933: Can you rephrase *semantic embedding* to something like *If False, the embeddings from the embedding model are used*? The input does not have to be semantic embeddings and can technically also be the input from Bag-of-Words/TF-IDF.

bertopic/_utils.py
```diff
@@ -147,3 +147,57 @@ def validate_distance_matrix(X, n_samples):
...
+        elif not isinstance(array, np.ndarray):
+            raise ValueError("The embeddings should be either a type of `numpy.ndarray` or `scipy.sparse.csr_matrix`")
+        return array
+
+    logger = MyLogger("WARNING")
+
+    if ctfidf_embeddings is None and semantic_embeddings is None:
+        raise ValueError("No topic representation was found.")
```
MaartenGr, line 183: Does this actually happen? Both representations are always saved, so I'm not sure whether it is necessary to check for this.

bertopic/_utils.py
```diff
@@ -147,3 +147,57 @@ def validate_distance_matrix(X, n_samples):
...
+    """
+
+    def to_ndarray(array: Union[np.ndarray, csr_matrix]) -> np.ndarray:
+        if isinstance(array, csr_matrix):
+            return array.toarray()
+        elif not isinstance(array, np.ndarray):
+            raise ValueError("The embeddings should be either a type of `numpy.ndarray` or `scipy.sparse.csr_matrix`")
+        return array
+
+    logger = MyLogger("WARNING")
```
MaartenGr, line 180: Just checking. If the logger is initialized outside of BERTopic, does the general logging still works? Even if you instantiate BERTopic multiple times?

bertopic/_utils.py
```diff
@@ -147,3 +147,57 @@ def validate_distance_matrix(X, n_samples):
...
+
+    if ctfidf_embeddings is None and semantic_embeddings is None:
+        raise ValueError("No topic representation was found.")
+
+    if ctfidf_embeddings is not None:
+        ctfidf_embeddings = to_ndarray(ctfidf_embeddings)
```
MaartenGr, line 186: Aside from casting not being necessary I think, you are already casting the embeddings to numpy even though it is not yet certain these embeddings are actually used.

MaartenGr (commented):
Thanks for your work on this! I left some comments that we can discuss as they relate to, for instance casting to numpy. Also, the tests fail on [python 3.8](https://github.com/MaartenGr/BERTopic/actions/runs/8738663864/job/23978590043?pr=1894) most likely as a result of these kinds of type hints not yet being supported.

bertopic/_utils.py
azikoss: Due to the code readability. Both embeddings can be `None` (by function types). This is the way how to say that one must not be None. 

Not a big deal, I can delete it if you like.

bertopic/_bertopic.py
azikoss: Sure. Done

bertopic/_utils.py
azikoss: Due to the code readability. Both embeddings can be `None` (by function types). This is the way how to say that one must not be None. 

Not a big deal, I can delete it if you like.

bertopic/_utils.py
azikoss: It depends who is calling this function, no? 

I can delete it though if you like.

bertopic/_utils.py
azikoss: Good point. I will have a look and update it.

bertopic/_utils.py
MaartenGr: > It depends who is calling this function, no?

What do you mean? Since this is a private function, nobody should be using in this in the first place right? Or is there some other usage where this might be happening?

bertopic/_utils.py
MaartenGr: As long as this function is not a public function, I think it is not necessary to implement this here. From a communicative perspective, it suggests that there are use cases where you have to check this. As a result, the error message should not only provide the problem but also the solution (or somewhere else, for instance in the docstrings). 

So I would only keep this if there are practical use cases where both might turn out to be `None`.

bertopic/_utils.py
azikoss: In my view, it would be less readable without the check

Here are two [best practices](https://betterprogramming.pub/clean-code-and-the-practice-of-programming-for-newcomers-2dfaeb0e31c3) that it breaks : 
Write for readability — code is read more often than it’s written.
Explicit is better than implicit.

Anyway, I will remove it.

bertopic/_utils.py
azikoss: My view is the same as above, but it is ok, I will remove it. 

The problem for me is that I did not see right away from the code (and perhaps this is my problem) that one representation always exists.

bertopic/_utils.py
azikoss: Done. Just to double-check. Semantic embeddings will always be `np.ndarray`, right?

bertopic/_utils.py
azikoss: Good point. Done.

bertopic/_utils.py
azikoss: Good catch. 

Since `getLogger(..)` returns just a reference, then there should no issue. The problem is with the repetitive setting of the level and possibly setting of other parameters. 

I would propose to create create a new method of MyLogger called `configure()` where the configuration takes place:
```
self.set_level(level)
self._add_handler()
self.logger.propagate = False
```

This would be called only once. 

Then, it would be possible to call anywhere MyLogger() and get the correct logger. What do you think?

bertopic/_utils.py
MaartenGr: In theory, whenever somebody uses a bag-of-words representations as the input for BERTopic, the "semantic embeddings" could also be sparse matrices. That's also why the term "semantic embeddings" might not be accurate if somebody uses a non-semantic representation as the input for BERTopic.

bertopic/_utils.py
MaartenGr: That sounds good considering the logger needs to be created only once and then retrieved a couple of times throughout the code.

bertopic/_utils.py
MaartenGr: Thanks for sharing! Do note though that I am not a newcomer as the blog post suggests 😉

I understand your view but do note that these best practices that you mention are very broad and can be interpreted from many perspectives. For instance, the code to me is less readable and implicitly communicates that we can certain situations arise from general use. 

For now, let's remove it considering the way the rest of the codebase follows a different style.

azikoss: I did not mean to suggest that you are a newcomer, sorry. I did not notice the title. I was looking for a reference to my approach. 

It is removed.

bertopic/_utils.py
azikoss: Ok. Done.

azikoss:
@MaartenGr pls let me know if there is anything else to address.

MaartenGr:
@azikoss Thanks for all the work so far! I was on holiday for the last two weeks and just getting back to a bunch of issues and PRs. Yours is high priority and I hope to get to it this weekend. Thanks for being patient with me.

MaartenGr:
I just did a first pass through the code and wanted to see whether all checks pass correctly but there seems to be a problem with the pipeline. To get this pipeline working for you, I would advise implementing the same fix as https://github.com/MaartenGr/BERTopic/pull/2008. That way, we can run the pipeline and see if everything works as intended.

azikoss:
@MaartenGr I am sorry for the delay. I just merged the latest changes. Is there anything else I should do with respect to https://github.com/MaartenGr/BERTopic/pull/2008 ?

Locally the tests are running successfully.

MaartenGr:
@azikoss I just merged the PR, congrats and thank you for all the hard work! Especially considering this PR started out small and grew into something bigger, I'm very grateful. Also, really cool feature to have as it will be helpful for instances where c-TF-IDF is not available (like after using `.merge_models`).

azikoss:
Great! My pleasure contributing to this great toolkit!

PR #1898, opened 2024-04-01
Fix issue with `.merge_models` seemingly skipping topic

MaartenGr:
Adresses #1866

In some cases, when using `.merge_models`, it seemed that the added topics to a base model were assigned with an incorrect numbering. For instance, if the maximum topic of the base model was 5, the newly added topics would start at 7. 

This issue fixes that problem and additionally ignores outlier classes as it seems there is no benefit to combining them. 

Lastly, this should also fix the topic embeddings being in the incorrect order.

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -3237,7 +3237,7 @@ class BERTopic:
 
             # Extract new topics
             new_topics = sorted([index - selected_topics["_outliers"] for index, sim in enumerate(sims) if sim < min_similarity])
-            max_topic = max(set(merged_topics["topics"])) + 1
+            max_topic = max(set(merged_topics["topics"]))
 
             # Merge Topic Representations
             new_topics_dict = {}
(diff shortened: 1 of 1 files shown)


bertopic/_bertopic.py
```diff
@@ -3237,41 +3237,41 @@ def merge_models(cls, models, min_similarity: float = .7, embedding_model=None):
 
             # Extract new topics
             new_topics = sorted([index - selected_topics["_outliers"] for index, sim in enumerate(sims) if sim < min_similarity])
-            max_topic = max(set(merged_topics["topics"]))
+            max_topic = max(set(merged_topics["topics"])) + 1
```
VinceMu, line 3240: Hi first off I just want to say that I love this library and that I appreciate all your hardwork on it.

I've patched my `merge_models()` function with the changes from this PR. I'm seeing a couple of topics with 'NaN' Names but non zero Count when I call`topic_info()` on the merged model. 
I think it may be due to `max_topic = max(set(merged_topics["topics"])) + 1`, should this just be `max_topic = max(set(merged_topics["topics"]))`?

MaartenGr: You are absolutely right! I had done that in my original code but it seems I forgot to transfer it over to this PR. A rather silly/lazy mistake. I just changed it and hopefully, it should work now.

VinceMu: Sorry forgot to say, works now - thank you!

MaartenGr: @VinceMu Awesome, thanks for checking!

Issue #1907, opened 2024-04-03
hierarchical_topics() produce incorrect output when three topics have the same distance


azikoss:
Hi there,

I have noticed that `hierarchical_topics(...)` method produces incorrect results when three or more topics have the same (tf-idf) distances. Let me illustrative it with an example.

```
from umap import UMAP
from bertopic import BERTopic

docs = (
    ["banana"] * 300 
    + ["banana apple"] * 300 
    + ["pear"] * 300 
    + ["lemon"] * 300 
    + ["clock"] * 300 
)

model = BERTopic(umap_model=UMAP(random_state=42))
topics, probs = model.fit_transform(docs)
hr = model.hierarchical_topics(docs)
hr
```
This outputs
![image](https://github.com/MaartenGr/BERTopic/assets/124801530/c403ae15-e974-48dd-83eb-6842cf2b7d6a)

The cluster with Parent_ID == 8 includes topics [1, 2, 3] but topic 3 is not mentioned in either the left or right child or their childs. 

**Why is it happening?** 
The [flat structure](https://github.com/MaartenGr/BERTopic/blob/master/bertopic/_bertopic.py#L1006) is created in each iteration. There is no guarantee that the new cluster will contain only two topics while the [code that follows](https://github.com/MaartenGr/BERTopic/blob/master/bertopic/_bertopic.py#L1031-L1041) presumes that.

**What should be the expected behavior?**
I think that a new cluster should emerge. Essentially, the structure should look like this

```
Parent_Id, Child_Left_ID, Child_Right_ID
     8               1               11
    11               2               3
```

MaartenGr:
Thank you for the reproducible example! This indeed seems to be an issue. Interestingly, I expect this would not happen often since distances are seldom the same right? How did you then stumble upon this issue? 

Either way, seeing as you already dove into the code, any suggestions how this could easily be resolved? Perhaps check for similar distances beforehand?

MaartenGr:
> A possible fix could be (deterministically) select a topic that has the same distance ith two other topics and add a tiny bit to the distance metric. 

If it's small enough that it's not significant, this would be the easiest solution without needing to change much to the source code. We could also add some (very slight perhaps uniform) noise to the entire distance metric such that all distances are affected equally. Either way, this sounds like a good approach as long as we can keep the noise small and logical. 

If you have the time, a PR would be appreciated! If not, I'll put it on my backlog but I'm not sure if I'm able to tackle this in the coming months.

[...]

PR #1929, opened 2024-04-17
Fix `hierarchical_topics(...)` when the distances between three clusters are the same

azikoss:
Adds functionality that makes sure that during calculating of `hierarchical_topics(...)` the distances between clusters are unique (by adding some small noise) otherwise the flatting of the hierarchy would produce incorrect values for "Topics" for these clusters (#[1907](https://github.com/MaartenGr/BERTopic/issues/1907))


bertopic/_utils.py
```diff
@@ -147,3 +147,29 @@ def validate_distance_matrix(X, n_samples):
...
+
+    Returns:
+         Unique distances sorted in the preserved increasing order.
+
+    Raises:
+          ValueError: If the distance array is not sorted in the increasing order.
+    """
+
+    if not np.all(np.diff(dists) >= 0):
+        raise ValueError("The distances must be sorted in the increasing order")
```
MaartenGr, line 169: When did you run into this issue that this is needed? I can imagine that even if the value error is given, this should not be the responsibility of the user, or should it?

bertopic/_utils.py
```diff
@@ -147,3 +147,29 @@ def validate_distance_matrix(X, n_samples):
         raise ValueError("Distance matrix cannot contain negative values.")
 
     return X
+
+
+def get_unique_distances(dists: np.array, noise_min=1e-3, noise_max=1e-1) -> np.array:
```
MaartenGr, line 152: What are the reasons for choosing these noise values? The max noise seems rather high still. Wouldn't that affect the actual distances too much?

Have you checked whether the noise only separates the identical distances whilst keeping the order in check? For instance, let's assume we have the following distances:

```python
distances = [0, 0.15, 0.2, 0.2, 0.3]
```

With the noise as is currently implemented, can we be 100% percent sure that the identical values (i.e., `0.2`) remain lower than the next highest value (i.e., `0.3`) but higher than the previous lower value (i.e., `0.15`)?
```

File: tests/test_utils.py [hunk 3]
```diff
@@ -32,3 +32,11 @@ def test_check_embeddings_shape():
...
                            [2, 3, 4]])
     check_embeddings_shape(embeddings, docs)
+
+
+def test_make_unique_distances():
+    dists = np.array([0, 0, 0.5, 0.75, 1, 1])
+    unique_dists = get_unique_distances(dists)
+
+    assert len(unique_dists) == len(dists), "The number of elements must be the same"
+    assert len(dists) == len(np.unique(unique_dists)), "The distances must be unique"
```
MaartenGr, line 42: As mentioned above, the noise might need to be smaller to keep the current order as this example shows large differences in distance between subsequent values. As a result, this test might be too lenient.

MaartenGr (commented):
Thanks for your work on this! I left a few notes with things that we might need to consider.

bertopic/_utils.py
azikoss: I was missing 0 at `1e-1`. 

1. I changed the defaults of the min.
2. Reimpleted the noise adding to make sure the ascending distance order is preserved.

bertopic/_utils.py
azikoss: I described the use case when this happens in the issue https://github.com/MaartenGr/BERTopic/issues/1907. 

I wanted to include this check as the assertion that the distances are originally sorted, otherwise the `get_unique_distances` would not work. if the distances are from linkage Z matrix, the order is always sorted.

What do you think about it?

MaartenGr: I'm not sure I understand. Issue #1907 describes when topics have the same distance not necessarily that the order of distances is descending instead of ascending. 

> I wanted to include this check as the assertion that the distances are originally sorted, otherwise the get_unique_distances would not work. if the distances are from linkage Z matrix, the order is always sorted.

Checks are always great but does it actually happen that the distances are not sorted? If yes, then how should the user use the error message to resolve the issue?

tests/test_utils.py
```diff
@@ -32,3 +32,18 @@ def test_check_embeddings_shape():
...
     check_embeddings_shape(embeddings, docs)
+
+
+def test_make_unique_distances():
+    def check_dists(dists: list[float], noise_max: float):
+        unique_dists = get_unique_distances(np.array(dists, dtype=float), noise_max=noise_max)
+        assert len(unique_dists) == len(dists), "The number of elements must be the same"
+        assert len(dists) == len(np.unique(unique_dists)), "The distances must be unique"
+
+    check_dists([0, 0, 0.5, 0.75, 1, 1], noise_max=1e-7)
```
MaartenGr, line 43: Have you checked the actual values of the updated distance list? When I run it, I get the following updated values:

```python
[0.00000000e+00 8.32483552e-08 5.00000000e-01 7.50000000e-01
 1.00000000e+00 2.00000008e+00]
```

The last value is twice as big which should not happen. I have a feeling the code for `get_unique_distances` could be simplified a bit. What about simply doing something like this:

```python
def get_unique_distances(dists):
    increment =  np.random.uniform(low=1e-5, high=1e-6)
    last_val = -float('inf')
    return [last_val := max(dist, last_val + increment) for dist in dists]

my_list = [0, 0, 0, 0.5, 0.75, 1, 1]
get_unique_distances(my_list)
```

tests/test_utils.py
azikoss: That was an issue in the code as mentioned in the comment below.

tests/test_utils.py
MaartenGr: > Are we ok with changing distances that do not have a duplicate?

Hmmm, my preference would indeed be to keep them as is as long as it requires no more than one or two lines of code. I would like to simplify this as much as possible.

azikoss: I simplified the code. Please have a look and let me know if you have any ideas.

MaartenGr: Thanks for the changes! I just tested it a bunch of times and it all looks good to me. Thanks for simplifying the code. I'll re-run the workflow to check whether everything passes. If it does, I will go ahead and merge the PR.

MaartenGr: The tests failed but I believe because you used `list[float]` which is not supported in python 3.8. Removing that should make the tests pass I think.

azikoss:
@MaartenGr could you please run the tests and see if we can merge this?

MaartenGr:
@azikoss After merging #1894 there are now a couple of small conflicts in this PR. Could you take a look? As soon as those are resolved I will go ahead and merge this PR.

MaartenGr:
Awesome, thank you for the work on this, it is greatly appreciated! 😄

PR #1941, opened 2024-04-22
ci: bump github actions versions

afuetterer:
Hi @MaartenGr,

this is my first PR to the project. I have read the contributor guide. I did not create an issue first, because the changes proposed are trivial. What do you think?

This PR bumps the versions of the used GitHub Actions, which will get rid of the warnings (e.g. `Node.js 16 actions are deprecated. [...]` in the [Annotations section under Actions](https://github.com/MaartenGr/BERTopic/actions/runs/8773439413).

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -18,9 +18,9 @@ jobs:
         python-version: [3.8, 3.9]
 
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


MaartenGr:
Since it is indeed a rather small change that does not affect functionality, no problem going straight to a PR. Just ran the new pipeline, no more warnings! Thanks for updating.

Issue #1950, opened 2024-04-26
Add support for Python 3.10+


afuetterer:
I tested `BERTopic` on my machine and in my fork using Python 3.12.

May I suggest adding 3.10, 3.11, 3.12 to `setup.py` and the GitHub actions CI job?

MaartenGr:
Although BERTopic does not limit the user to a maximum version of Python (so that's already supported), there are indeed no tests available for those higher versions. So yes, adding them would be great! I believe there might be some specific issues on higher python versions with specific package versions but the tests cannot cover all these cases unfortunately.

afuetterer:
Accidentally closed.

I will add a new PR then, about this in the upcoming days. As I said I tested with Python 3.12 and the tests were passing.

PR #1968, opened 2024-05-03
feat: add support for python 3.10+

afuetterer:
See #1950

This will run the test suite five times for Python 3.8-3.12. You could also opt for only running 3.8 and 3.12. Running it five times is safer I guess, so you will know if things go wrong in 3.10 and 3.11 as well.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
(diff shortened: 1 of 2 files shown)


MaartenGr:
Great! Thanks for the PR. Considering these all run in parallel, it's fine by me to have them all run. It definitely helps with tracking compatibility. A bit surprising though that it all passed without any issues since there have been some reports I believe of certain versions not working in specific use cases but that might be a problem with the testing suite. 

Either way, thanks!

afuetterer:
Yes, maybe there are edge cases, that are not covered by the test suite.

Issue #1969, opened 2024-05-04
Switch from setup.py to pyproject.toml


afuetterer:
The `pyproject.toml` file is decribed in [PEP 621](https://peps.python.org/pep-0621/). It is the recommended way to package Python projects these days.

Please have a look at my fork's version of this file: <https://github.com/afuetterer/BERTopic/blob/1969-toml/pyproject.toml>

You are still able to do a local `pip install .` or `pip install bertopic[gensim]` from pypi. 

Would you accept a PR to convert BERTopic to use pyproject.toml? Please let me know what you think.

Refs:
- https://packaging.python.org/en/latest/tutorials/packaging-projects/
- https://packaging.python.org/en/latest/guides/modernize-setup-py-project/
- https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

MaartenGr:
Thanks for the suggestion! What would you think is the main added benefit of doing so in the context of this package? Also, I remember it did not support editable installs a while back. Any idea if that's still the case?

afuetterer:
First of all: Editable installs are possible, I use them all the time.

I think the added benefits would be:

- easier definition of entrypoints, e.g. CLI
- ability to add additional tooling config in the same file, so the complete project configuration can live in one place
- easier definition of (nested) dependency groups
- ability to use Dependabot for getting scheduled dependency updates (also possible with requirements.txt, but not setup.py)

From llama3's point of view:
- Consistent configuration: pyproject.toml is a TOML file that can be used to configure your package, including its name, version, and dependencies. This provides a consistent way to manage package metadata.
-  Improved readability: The TOML format is more readable than the Python code in setup.py. This makes it easier for developers to understand the configuration of their packages.
- Better support for complex configurations: pyproject.toml supports more complex configurations, such as conditional dependencies and custom build steps. This allows you to manage more complex package dependencies and builds.
- Integration with other tools: pyproject.toml is used by other tools, such as Poetry and pipenv, to manage Python packages. This provides a consistent way to manage packages across different development environments.
- Improved security: The use of TOML files for configuration reduces the risk of code injection attacks that can occur when using Python code in setup.py.

What do you think? Should I submit a PR?

MaartenGr:
> First of all: Editable installs are possible, I use them all the time.

[...]

PR #1978, opened 2024-05-07
build: switch from setup.py to pyproject.toml

afuetterer:
See #1969


Makefile
```diff
@@ -14,8 +14,7 @@ docs:
 	mkdocs serve
 
 pypi:
-	python setup.py sdist
-	python setup.py bdist_wheel --universal
+	python -m build
```
afuetterer, line 17: I have overseen this command in the Makefile before.

The pypi command used the deleted setup.py file, which lead to an error.

I added `python -m build` instead, which uses the new recommended [build](https://build.pypa.io) tool.

afuetterer:
What do you think?

MaartenGr:
Thanks for the PR and the ping! There are several PRs/issues/discussions open that require my attention so it might take a while before I can explore this PR.

afuetterer:
Of course, sorry. :)

Makefile
MaartenGr: If I'm not mistaken, using `build` would require a new [pipeline](https://carpentries-incubator.github.io/python_packaging/instructor/05-publishing.html#build-and-twine) which complicates my process to push it to pypi. Instead, I would suggest following the [scikit-lego](https://github.com/koaning/scikit-lego/blob/main/setup.py) approach such that the original pipeline would remain intact.

pyproject.toml
```diff
@@ -0,0 +1,101 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
MaartenGr, line 3: This needs to be setuptools if we are indeed going back to the original approach of pushing to pypi.

pyproject.toml
```diff
@@ -0,0 +1,101 @@
...
+
+[project]
+name = "bertopic"
+dynamic = ["version"]
```
MaartenGr, line 7: I would prefer an explicit version akin to:

```
version = "0.16.2"
```

pyproject.toml
```diff
@@ -0,0 +1,101 @@
...
+description = "BERTopic performs topic Modeling with state-of-the-art transformer models."
+readme = "README.md"
+license = {text = "MIT"}
```
MaartenGr, line 10:
```suggestion
license = {file = "LICENSE"}
```

pyproject.toml
```diff
@@ -0,0 +1,101 @@
...
+]
+vision = [
+    "accelerate>=0.19.0", # To prevent "cannot import name 'PartialState' from 'accelerate'"
+    "Pillow>=9.2.0",
+]
+
+[project.urls]
+Documentation = "https://maartengr.github.io/BERTopic/"
+Homepage = "https://github.com/MaartenGr/BERTopic"
+"Issue Tracker" = "https://github.com/MaartenGr/BERTopic/issues"
```
MaartenGr, line 92:
```suggestion
Issues = "https://github.com/MaartenGr/BERTopic/issues"
```

pyproject.toml
```diff
@@ -0,0 +1,101 @@
...
+"Source Code" = "https://github.com/MaartenGr/BERTopic/"
```
MaartenGr, line 93:
```suggestion
Repository = "https://github.com/MaartenGr/BERTopic.git"
```

pyproject.toml
```diff
@@ -0,0 +1,101 @@
...
+
+[tool.hatch.version]
+path = "bertopic/__init__.py"
+
+[tool.hatch.build.targets.sdist]
+include = [
+    "/bertopic",
+]
```
MaartenGr, line 101: If setuptools, something like this:

```
[tool.setuptools.packages.find]
include = ["bertopic*"]
exclude = [
    "docs",
    "images",
    "notebooks",
    "tests",
]
```

MaartenGr (commented):
Left a couple of questions/suggestions that we can go over to make sure the general flow of installation remains the same.

afuetterer:
I will adapt soon.

pyproject.toml
afuetterer: You want to track the version information only in the pyproject.toml then or duplicate it in the src as well? I recommend keeping a single source of truth.

If the version is set in the toml a common pattern is to use:

```python
from importlib.metadata import version
__version__ = version("bertopic")
```

In case you need the version at runtime.

Makefile
afuetterer: I'm sorry, but what exactly do you mean by pipeline?
In my opinion the result of the two replaced lines should be identical to running python -m build. Running `make pypi` should build the sdist and wheel in `dist`, which gets picked up by twine.
I don't think the setup.py is needed any longer. But sure, if you prefer to keep it, the minimal one from your link should be fine.

What do you think?

pyproject.toml
afuetterer: Sure, you prefer setuptools? The build backends (setuptools, hatch, flit, pdm,...) should work interchangeably.
I chose hatch because pypa recommends it in the packaging tutorial.
Actually you can also do hatch build and hatch publish.

Makefile
MaartenGr: You know what, let's indeed remove `setup.py` but keep setuptools as the build system. That removes an additional unnecessary file but still allows me to keep the familiarity I have with setuptools. 

Makes debugging for me easier when I do stuff with pypi but still allows for the newer format. Does that make sense?

pyproject.toml
MaartenGr: Yep, it's also a manner of convenience since I'm familiar with it and will be the one debugging things when pushing to pypi.

pyproject.toml
afuetterer: docs, images, notebooks are not picked up by setuptools, so I did not include those.

MaartenGr:
Awesome, thank you for the work on this! LGTM 😄

PR #1984, opened 2024-05-10
Make sklearn embedding backend auto-select more cautious

### LINKED ISSUE #1980: Warn when automatically choosing SklearnEmbedder backend
I recently installed an older version of BERTopic which caused an incompatible version of sentence-transformers and huggingface_hub to be installed. With these versions, [this error](https://github.com/huggingface/huggingface_hub/issues/1199) is generated, but if you rely on the default model being used, it will be caught by BERTopic, and the default sklearn embedder will be selected completely silently.

I was wondering why the results were so bad until I tried manually importing SentenceTransformers to create the model myself, which then gave me the ModuleNotFoundError. Had I not tried this, I would've had no idea I was using the SklearnEmbedder.

This doesn't happen if you install the current version of BERTopic in a fresh environment, but it's entirely possible that a similar incompatibility could occur in the future, other package constraints could cause an older version of sentence-transformers to be installed, or a broken environment will cause a ModuleNotFoundError even if all the packages are supposed to be installed.

I think it would be wise to:
- find an alternative way of more specifically detecting minimal installs (e.g. the entire SentenceTransformers package being missing), to prevent this code path activating unintentionally
- provide a warning if sklearn is selected automatically. The fact that the default model is `all-MiniLM-L6-v2` is mentioned a few times in the docs, so most users will be expecting it and have no idea they could get a Tfidf/SVD model instead. This may negatively affect users of the minimal install who rely on the default TfidfVectorizer/TruncatedSVD model, so maybe a special model name could be added to select this model without a warning?

Of course this wouldn't prevent issues with older versions of BERTopic, but it would reduce the risk of this happening in future.

freddyheppell:
Fixes #1980 

- Checks which module the error ocurred in before selecting an sklearn backend by default. If it was `sentence_transformers`, select skelarn as it's probably a minimal install. If any other module, re-raise as this is abnormal.
- Logs an INFO message when the sklearn backend is selected, if verbose is enabled
- Updates the comments on `select_backend` as I think they were quite out of date

I ended up instantiating a new logger for this as it felt better to pass `verbose` than the whole logger.

### DIFF
--- a/bertopic/backend/_utils.py
+++ b/bertopic/backend/_utils.py
@@ -7,6 +7,9 @@ from sklearn.decomposition import TruncatedSVD
 from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.pipeline import Pipeline as ScikitPipeline
 
+from bertopic._utils import MyLogger
+
+logger = MyLogger("WARNING")
 
 languages = [
     "arabic",
@@ -66,14 +69,20 @@ languages = [
 
 
 def select_backend(embedding_model,
-                   language: str = None) -> BaseEmbedder:
-    """ Select an embedding model based on language or a specific sentence transformer models.
+                   language: str = None,
+                   verbose: bool = False) -> BaseEmbedder:
+    """ Select an embedding model based on language or a specific provided model.
     When selecting a language, we choose all-MiniLM-L6-v2 for English and
     paraphrase-multilingual-MiniLM-L12-v2 for all other languages as it support 100+ languages.
+    If sentence-transformers is not installed, in the case of a lightweight installation,
+    a scikit-learn backend is default.
 
     Returns:
-        model: Either a Sentence-Transformer or Flair model
+        model: The selected model backend.
     """
+
+    logger.set_level("INFO" if verbose else "WARNING")
+
     # BERTopic language backend
     if isinstance(embedding_model, BaseEmbedder):
         return embedding_model
@@ -126,8 +135,14 @@ def select_backend(embedding_model,
                                 "Else, please select a language from the following list:\n"
                                 f"{languages}")
 
-        # Only for light-weight installation
-        except ModuleNotFoundError:
+        # A ModuleNotFoundError might be a lightweight installation
+        except ModuleNotFoundError as e:
+            if e.name != "sentence_transformers":
+                # Error occurred in a downstream module, probably not a lightweight install
+                raise e
+            # Whole sentence_transformers module is missing, probably a lightweight install
+            if verbose:
+                logger.info("Automatically selecting lightweight scikit-learn embedding backend as sentence-transformers appears to not be installed.")
             pipe = make_pipeline(TfidfVectorizer(), TruncatedSVD(100))
             return SklearnEmbedder(pipe)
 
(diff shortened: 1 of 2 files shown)


bertopic/backend/_utils.py
```diff
@@ -126,8 +134,14 @@ def select_backend(embedding_model,
...
-        # Only for light-weight installation
-        except ModuleNotFoundError:
+        # A ModuleNotFoundError might be a lightweight installation
+        except ModuleNotFoundError as e:
+            if e.name != "sentence_transformers":
+                # Error occurred in a downstream module, probably not a lightweight install
+                raise e
+            # Whole sentence_transformers module is missing, probably a lightweight install
+            if verbose:
+                logger.info("Automatically selecting lightweight scikit-learn embedding backend as sentence-transformers appears to not be installed.")
```
MaartenGr, line 144: The if statement is not needed since you already set the logging level right?

bertopic/backend/_utils.py
```diff
@@ -2,11 +2,13 @@
 
 # Imports for light-weight variant of BERTopic
 from bertopic.backend._sklearn import SklearnEmbedder
+from bertopic._utils import MyLogger
 from sklearn.pipeline import make_pipeline
 from sklearn.decomposition import TruncatedSVD
 from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.pipeline import Pipeline as ScikitPipeline
 
+logger = MyLogger("WARNING")
```
MaartenGr, line 11: This might create duplicates of the logger but I am not entirely sure. I believe this might be related: https://github.com/MaartenGr/BERTopic/pull/1894#discussion_r1574244162

MaartenGr (commented):
Thanks for the PR! I had a few questions/suggestions but other than that it looks good. There is currently an issue with the pipeline as a result of the scikit-learn v1.5 upgrade. To get this pipeline working for you, I would advise implementing the same fix as https://github.com/MaartenGr/BERTopic/pull/2008. That way, we can run the pipeline and see if everything works as intended.

bertopic/backend/_utils.py
freddyheppell: I think as a comment there says, instantiating MyLogger will return a reference to the same logger instance each time. Although setting the log level here will affect all instances of MyLogger (because they're the same underlying logger), because it's also in the `__init__()` of `BERTopic` it should then set it correctly. But the suggestions to clean this up in that PR sound sensible, I'm happy to adopt those once it's merged.

freddyheppell: Oh but then I've added setting the level in `select_backend()` because I didn't realise the implications of `MyLogger` at that point. I think I can just remove the line I added to do that, and it'll be fine because the log level is already set.

bertopic/backend/_utils.py
freddyheppell: I think I only did this because that's how it's done in a few other places e.g. `BERTopic._extract_topics()`. If that's unnecessary I can remove

freddyheppell:
Just rebased this branch onto the latest main so it has the #2008 fix now

bertopic/backend/_utils.py
MaartenGr: For now this is fine. I might update the logger at some point but that will be separate from this PR.

bertopic/backend/_utils.py
MaartenGr: Ah, you're right! Let's keep it for now and when the logger updates I might change it.

MaartenGr:
@freddyheppell Thanks for your patience concerning the pipeline fix. Everything looks good to me so I'll go ahead and merge it. Thanks for your contribution to the repo! 😄

Issue #1979, opened 2024-05-07
Proposal: Improve error output for OpenAI resource


steven-solomon:
# Scenario

When utilizing OpenAI in an enterprise context, there are additional policy and RAI compliance rules applied to the instance which need to be handled. This will enable more folks to start using BERTopic within Azure for their organization.

# High Level Proposal

I would like to modify `representation.OpenAI` to output more detail about errors returned from [AzureOpenAI](https://github.com/openai/openai-python/blob/89478ce06e2161fda416c685b774a97a041ae185/src/openai/lib/azure.py#L64), which is already part of the `openai` package. 

Additionally, I think it logging document indices used as part of a representation would make debugging error scenarios in this API easier to fix.

Before moving forwards with development, I await your feedback.

MaartenGr:
Thanks for sharing this issue! Just to be sure I understand correctly, you want to implement a way to handle most `finish_reason` as a way for the user to understand why the output was empty/truncated/missing, right? 

Sounds great! Additional information during inference would be more than welcome considering users have struggled with missing output in the past. 

> Handle all four [CompletionChoice.finish_reasons](https://platform.openai.com/docs/guides/text-generation/chat-completions-api)
> Open question: How do you want to handle the length and function_call and null finish_reasons

I don't think at the moment we need to do anything with `function_call` since BERTopic does not make use of it and the models generally follow instructions quite well. 

I believe `length` can be logged similar to `content_filter`, since both do something with the output. Here, we can mention to use the truncation options available in BERTopic to prevent these issues. With `length`, we might need to add "incomplete output due to..." or something similar to the label if it happens to be truncated. If it is empty, we can either leave it empty and log it or create a label that says something like "incomple output due to...". 

It seems that `null` should only happen when you try to access the API when it is already running which should generally not happen unless the user was already running some process right? Having said that, logging it like `length` and `content_filter` seems like a possible solution.

> Log repr_doc_ids which caused sensitive responses

Agreed, and since BERTopic does not pass all documents to the API I would not expect excessive logging. 

All in all, sounds good. Looking forward to this!

steven-solomon:
> Thanks for sharing this issue! Just to be sure I understand correctly, you want to implement a way to handle most finish_reason as a way for the user to understand why the output was empty/truncated/missing, right?

Yes, that is correct.

@MaartenGr, thanks for your feedback

My plan will be as follows:
- `finish_reason` modifications
[...]

PR #1985, opened 2024-05-10
Extend OpenAI `finish_reason` handling

soonernotfaster:
As part of https://github.com/MaartenGr/BERTopic/issues/1979, I added:

- explicit handling for the `finish_reason` values: `length`, `content_filter`, and `stop`
- handling`finish_reason` for any unspecified cases
- log `repr_doc_ids` in all error cases to improve debugging

Fixes https://github.com/MaartenGr/BERTopic/issues/1979


bertopic/representation/_openai.py
```diff
@@ -217,11 +218,21 @@ def extract_topics(self,
...
-                # Adresses #1570 for potential issues with OpenAI's content filter
-                if hasattr(response.choices[0].message, "content"):
-                    label = response.choices[0].message.content.strip().replace("topic: ", "")
+                choice = response.choices[0]
+
+                if choice.finish_reason == "stop":
+                    label = choice.message.content.strip().replace("topic: ", "")
+                elif choice.finish_reason == "length":
+                    logger.warn(f"Extracing Topics - Length limit reached for doc_ids ({repr_doc_ids})")
+                    if hasattr(response.choices[0].message, "content"):
```
steven-solomon, line 227: I am not sure of the APIs behavior when hitting the token limit, so I kept the check for `content`.

bertopic/representation/_openai.py
```diff
@@ -217,11 +218,21 @@ def extract_topics(self,
...
+                        label = choice.message.content.strip().replace("topic: ", "")
+                    else:
+                        label = "Incomple output due to token limit being reached"
+                elif choice.finish_reason == "content_filter":
+                    logger.warn(f"Extracing Topics - Content filtered for doc_ids ({repr_doc_ids})")
+                    label = "Output content filtered by OpenAI"
                 else:
+                    logger.warn(f"Extracing Topics - No label due to finish_reason {choice.finish_reason} for doc_ids ({repr_doc_ids})")
```
steven-solomon, line 235: I am not in love with exposing the `finish_reason` concept in the error message but I wonder with folks will have to learn about it when debugging the issue anyway. Wyt?

steven-solomon: I also think I need to index into `repr_doc_ids` based on the index. Kudos to @sdehm for pointing that out.

bertopic/representation/_openai.py
```diff
@@ -5,8 +5,9 @@
 from scipy.sparse import csr_matrix
 from typing import Mapping, List, Tuple, Any, Union, Callable
 from bertopic.representation._base import BaseRepresentation
-from bertopic.representation._utils import retry_with_exponential_backoff, truncate_document
+from bertopic.representation._utils import retry_with_exponential_backoff, truncate_document, MyLogger
 
+logger = MyLogger("WARNING")
```
MaartenGr, line 10: I'm not sure whether importing the logger will work with the current logging procedure. My guess is that it will create a separate logger which does things twice. Also, the model should control the verbosity so that this warning is suppressed if needed by the user. 

I believe, but I will have to check, that the functionality of the logger will change in an open PR.

bertopic/representation/_openai.py
```diff
@@ -37,7 +38,7 @@
 Topic name:"""
 
 DEFAULT_CHAT_PROMPT = """
-I have a topic that contains the following documents: 
+I have a topic that contains the following documents:
```
MaartenGr, line 41: Although I expect no issues, it would be interesting to see the effect of removing a `\n` from this prompt. In my experience with recent LLMs (Phi-3/Llama-3), this actually can have a decent influence on performance (especially with respect to following instructions). Either, good change!

bertopic/representation/_openai.py
MaartenGr: For now, this is fine I guess. Also, you can replace `response.choices[0]` with `choice`. Also, and this might be just me, could we replace the variable name `choice` with `output`? I find the term `choice` not easily readable/explicit.

bertopic/representation/_openai.py
```diff
@@ -217,11 +218,21 @@ def extract_topics(self,
...
+                choice = response.choices[0]
+
+                if choice.finish_reason == "stop":
+                    label = choice.message.content.strip().replace("topic: ", "")
+                elif choice.finish_reason == "length":
+                    logger.warn(f"Extracing Topics - Length limit reached for doc_ids ({repr_doc_ids})")
+                    if hasattr(response.choices[0].message, "content"):
+                        label = choice.message.content.strip().replace("topic: ", "")
+                    else:
+                        label = "Incomple output due to token limit being reached"
```
MaartenGr, line 230: Should be "Incomplete" instead of "Incomple"

bertopic/representation/_openai.py
```diff
@@ -217,11 +218,21 @@ def extract_topics(self,
...
+                elif choice.finish_reason == "content_filter":
+                    logger.warn(f"Extracing Topics - Content filtered for doc_ids ({repr_doc_ids})")
```
MaartenGr, line 232: Perhaps make the warning a bit more explicit? For instance, "The content filter of OpenAI was trigger for the following documents IDs:"

bertopic/representation/_openai.py
MaartenGr: > I am not in love with exposing the finish_reason concept in the error message but I wonder whether folks will have to learn about it when debugging the issue anyway. Wyt?

If that reason didn't trigger any of the other if statements, then to me it makes sense to log it since we didn't expect it and there is no reason for the user to know which one it is. 

Also, perhaps make the warning a bit more explicit. For instance, "OpenAI Topic Representation - Couldn't create a label due to {choice.finish_reason} for the following document IDs"

I think being explicit here might improve any potential debugging.

MaartenGr (commented):
Apologies for the delay, I left some comments here and there.

bertopic/representation/_openai.py
steven-solomon: That makes total sense. Since we want to leave verbosity controls to the model, I could use the logger instance from `topic_model.logger`. Wyt?

bertopic/representation/_openai.py
steven-solomon: That's a great point. I am tempted to undo this change then I don't necessarily want to modify any behavior other than logging.

steven-solomon:
@MaartenGr, I made the following updates based on your feedback:

- Renamed `choice` to `output`
- Introduce a logging prefix `OpenAI Topic Representation` 
- Improve copy of logs based on your feedback
  - Modified references to `doc_ids` to use `document IDs` 
  - Fixed typos
  
 Open questions:
- [Would you like me to grab the logger off of the `topic_model`?](https://github.com/MaartenGr/BERTopic/pull/1985#discussion_r1610255265)
- [Should I defer any changes to the prompt?](https://github.com/MaartenGr/BERTopic/pull/1985#discussion_r1610256122)

bertopic/representation/_openai.py
MaartenGr: Good thinking, that would things much easier.

bertopic/representation/_openai.py
MaartenGr: Let's keep it for now and potentially change this in the future.

MaartenGr:
@steven-solomon Apologies for the delay, last few weeks were busy. It seems that there are some conflicts with the main branch. Other than that, I just ran the pipeline but expect this to run without any issues. Moreover, it seems there might be an issue with selecting the correct `output` variable.

steven-solomon:
> Moreover, it seems there might be an issue with selecting the correct output variable.

I think I see what you mean. The definition of `output` on L221 should be up another level so it is accessible in L243. Is that correct or is there something else amiss? 

I am planning on adding the change to use the model's logger in the next day or two.

MaartenGr:
@steven-solomon I think I see what you mean. The definition of output on L221 should be up another level so it is accessible in L243. Is that correct or is there something else amiss?

No, I meant that the output variable `response` on L242 and L240 are not doing anything:

```python
                if self.exponential_backoff:
                    response = completions_with_backoff(self.client, model=self.model, prompt=prompt, **self.generator_kwargs)
                else:
                    response = self.client.completions.create(model=self.model, prompt=prompt, **self.generator_kwargs)
```

since they are not accessed on L243:

```python
                label = output.text.strip()
```

steven-solomon:
@MaartenGr I have made the switch the pull the logger from the `topic_model` parameter and added a fix for L243.

MaartenGr:
@steven-solomon There is no `topic_model.logger` attribute so that wouldn't work, unfortunately. What was wrong with the previous approach?

soonernotfaster:
Any objection to me closing this out?

MaartenGr:
@soonernotfaster No problem if you want to close this!

Issue #1981, opened 2024-05-08
PartOfSpeech representation reproducibility and word with index 0


Greenpp:
Hi Maarten

### Part 1
While working with BERTopic, I encountered a problem with reproducibility of representations. I made sure to set `random_state` wherever possible. After reviewing all similar issues and trying things like disabling MST approximation in HDBSCAN (`approx_min_span_tree=False`) or setting global random state with numpy (`numpy.random.seed`), I started digging into the library. 

I found that the values remained constant until the `PartOfSpeech` representation module and switching it to another one resolved the issue. The problem appears to be initially caused by the deduplication method (`list(set())`) used at lines [121](https://github.com/MaartenGr/BERTopic/blob/9e5558bc2d04797280b10576added63730459621/bertopic/representation/_pos.py#L121) and [130](https://github.com/MaartenGr/BERTopic/blob/9e5558bc2d04797280b10576added63730459621/bertopic/representation/_pos.py#L130). Because the `hash` function used for generating set keys is seeded at the start of the interpreter (seed can be overridden using `PYTHONHASHSEED` env variable), the output of such deduplication is different with each run. This behavior causes `word_indices` at line [144](https://github.com/MaartenGr/BERTopic/blob/9e5558bc2d04797280b10576added63730459621/bertopic/representation/_pos.py#L144) to change with each run. Which is later problematic when sorting keywords with the same c-TF-IDF as they are arranged differently.

### Part 2
When looking at the PoS code, I noticed that `word_indices` at line [144](https://github.com/MaartenGr/BERTopic/blob/9e5558bc2d04797280b10576added63730459621/bertopic/representation/_pos.py#L144) are generated using the following condition `if words_lookup.get(keyword)` which ignores the first word returned by `get_feature_names_out`. It looks like an error.

## MRE

```python
from bertopic import BERTopic
from bertopic.representation import PartOfSpeech
from sklearn.datasets import fetch_20newsgroups
from sklearn.feature_extraction.text import CountVectorizer
from umap import UMAP

docs = fetch_20newsgroups(subset='all',  remove=('headers', 'footers', 'quotes'), random_state=42)['data'][:100]

umap_model = UMAP(n_neighbors=15, n_components=5, min_dist=0.0, metric='cosine', random_state=42)
vectorizer_model = CountVectorizer(stop_words="english", ngram_range=(1, 2))
representation_model = PartOfSpeech()
topic_model = BERTopic(
    umap_model=umap_model,
[...]

PR #1996, opened 2024-05-18
Fixed reproducibility problem from issue #1981

Greenpp:
Fixed PartOfSpeech representation reproducibility problem described in issued #1981

### DIFF
--- a/bertopic/representation/_pos.py
+++ b/bertopic/representation/_pos.py
@@ -141,8 +141,8 @@ class PartOfSpeech(BaseRepresentation):
         updated_topics = {topic: [] for topic in topics.keys()}
 
         for topic, candidate_keywords in candidate_topics.items():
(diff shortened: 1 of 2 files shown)


MaartenGr:
Awesome, thanks for the solution and PR! The pipeline is failing which I do [not think is related](https://github.com/MaartenGr/BERTopic/issues/2005) to this PR. Having said that, I'll check it out in the coming days to see if I can fix it so that we can properly run this.

MaartenGr:
To get this pipeline working for you, I would advise implementing the same fix as https://github.com/MaartenGr/BERTopic/pull/2008. That way, we can run the pipeline and see if everything works as intended.

MaartenGr:
Awesome, the pipeline now works well. Thanks for your help on this!

Issue #1846, opened 2024-02-28
Error when setting chain representation models as main


coryzhangia:
When chain models are named anything else, it works fine. But if I want to use a chain model as the main representation, it will produce an error.

```python
representation_model = {
    "Main": [KeyBERT, MMR],
#    "ChatGPT": aspect_model1
   "KeyBERT": KeyBERT,
   "BERT MMR": [KeyBERT, MMR],
    "POS": aspect_model2,
    "POS MMR": [aspect_model2, MMR]
}``` 

#The error produced:

```python

File ~/.local/lib/python3.10/site-packages/bertopic/_bertopic.py:433, in BERTopic.fit_transform(self, documents, embeddings, images, y)
    430     self._save_representative_docs(custom_documents)
    431 else:
    432     # Extract topics by calculating c-TF-IDF
--> 433     self._extract_topics(documents, embeddings=embeddings, verbose=self.verbose)
    435     # Reduce topics
    436     if self.nr_topics:

File ~/.local/lib/python3.10/site-packages/bertopic/_bertopic.py:3637, in BERTopic._extract_topics(self, documents, embeddings, mappings, verbose)
   3635 documents_per_topic = documents.groupby(['Topic'], as_index=False).agg({'Document': ' '.join})
   3636 self.c_tf_idf_, words = self._c_tf_idf(documents_per_topic)
-> 3637 self.topic_representations_ = self._extract_words_per_topic(words, documents)
   3638 self._create_topic_vectors(documents=documents, embeddings=embeddings, mappings=mappings)
   3639 self.topic_labels_ = {key: f"{key}_" + "_".join([word[0] for word in values[:4]])
   3640                       for key, values in
   3641                       self.topic_representations_.items()}

File ~/.local/lib/python3.10/site-packages/bertopic/_bertopic.py:3925, in BERTopic._extract_words_per_topic(self, words, documents, c_tf_idf, calculate_aspects)
   3923 elif isinstance(self.representation_model, dict):
   3924     if self.representation_model.get("Main"):
-> 3925         topics = self.representation_model["Main"].extract_topics(self, documents, c_tf_idf, topics)
   3926 topics = {label: values[:self.top_n_words] for label, values in topics.items()}
   3928 # Extract additional topic aspects

AttributeError: 'list' object has no attribute 'extract_topics'
```

MaartenGr:
[...]

PR #2002, opened 2024-05-22
In multi-aspect context, allow Main model to be chained [Issue #1846]

ddicato:


bertopic/_bertopic.py
```diff
@@ -4071,7 +4071,11 @@ def _extract_words_per_topic(self,
             topics = self.representation_model.extract_topics(self, documents, c_tf_idf, topics)
         elif isinstance(self.representation_model, dict):
             if self.representation_model.get("Main"):
-                topics = self.representation_model["Main"].extract_topics(self, documents, c_tf_idf, topics)
+                if isinstance(self.representation_model["Main"], BaseRepresentation):
+                    topics = self.representation_model["Main"].extract_topics(self, documents, c_tf_idf, topics)
+                elif isinstance(self.representation_model["Main"], list):
+                    for tuner in self.representation_model["Main"]:
+                        topics = tuner.extract_topics(self, documents, c_tf_idf, topics)
         topics = {label: values[:self.top_n_words] for label, values in topics.items()}
```
MaartenGr, line 4079: Although it would be nice to have chained main models, it does bring a number of difficulties.

First, whenever you chain the main representation, any additional aspects that you want to create are then created on top of that chained representation. This inhibits the functionality of creating multiple aspects. Thus, whenever you use a chained Main representation, you cannot really use additional aspects as they will use the same chained base as the Main representation. 

Second, whenever you create a chained representation as the Main model, it will still apply `top_n_words`, which might not be appropriate depending on the last chain the representation.

ddicato: Interesting, thanks for explaining that. I clearly wasn't aware of the full implications of this change.

For the first issue: What is the desired behavior here? If I'm creating multiple aspects, I would think that each chain would be independent, not concatenated onto `"Main"`. If that's the intended behavior, I think I can fix that.

For the second issue: It seems like this is a problem with chained main representations in general, whether or not we have multiple aspects. What is the intended behavior of `top_n_words`? Do we wish to automatically apply it after `"Main"` and before every non-Main aspect? Or do we wish to apply it after every non-Main aspect? Or do we prefer allowing the user some control on if/when `top_n_words` is applied to each aspect?

MaartenGr: > For the first issue: What is the desired behavior here? If I'm creating multiple aspects, I would think that each chain would be independent, not concatenated onto "Main". If that's the intended behavior, I think I can fix that.

The desired behavior would be that any additional aspect is built on top of the default "Main" representation which is the c-TF-IDF representation. If it would be built on top of the chained "Main" representation, then that would create strange results, especially from a user-experience perspective. 

> For the second issue: It seems like this is a problem with chained main representations in general, whether or not we have multiple aspects. What is the intended behavior of top_n_words? Do we wish to automatically apply it after "Main" and before every non-Main aspect? Or do we wish to apply it after every non-Main aspect? Or do we prefer allowing the user to control whether/when top_n_words is applied to each aspect?

`top_n_words` is used mainly for c-TF-IDF, so the default representation. Other representation models might have their own `top_n_words` parameter.

ddicato: Thanks for the feedback. I've pushed a new version where I attempted the following:

For Main representation:
- If Main representation is missing, None, or falsy, then use default c-tf-idf + top_n_words
- Avoid applying top_n_words in all other cases
- Process `BaseRepresentation`, `list[BaseRepresentation]`, and `dict[str, BaseRepresentation | list[BaseRepresentation]]`
- raise `TypeError` in all other cases

For other aspects:
- Copy c-tf-idf topics directly to use as input
- If `aspect_model` is falsy, then use default c-tf-idf + top_n_words
- Avoid applying top_n_words in all other cases
- Process `BaseRepresentation` and `list[BaseRepresentation]`
- raise `TypeError` in all other cases

MaartenGr:
@ddicato Apologies for the late reply and thank you for the changes! This is exactly the behavior as intended and it is nicely and elegantly coded.

PR #2004, opened 2024-05-22
None Object is returned on .learn_one

Proteusiq:
removed the assignment as it would self.model to None. self.model.predict_one returns None. Just like dictionary update. So assignment kill the object.


MaartenGr:
Ah, that should indeed be changed. No comments here, LGTM!

MaartenGr:
@hamedbh Ah right, I forgot to update that documentation. I just did and it should now be [correct](https://maartengr.github.io/BERTopic/getting_started/online/online.html#river).

hamedbh:
@MaartenGr that's great, thank you!

PR #2039, opened 2024-06-06
Fix calls to self.verbose in class methods

freddyheppell:
In #1984 a few usages of `self.verbose` in class methods slipped by.

I've changed these to check whether the model (or any model in the case of `merge_models`) is set to be verbose. I did consider just always running them as non-verbose, but I thought that really if you're loading an existing model, BERTopic deciding to use a different model by itself is even worse than usual.

Of course this also indicates these code paths don't have test coverage as this would've failed otherwise. Not sure if that should also be addressed.

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -3146,7 +3146,7 @@ class BERTopic:
             with open(file_or_dir, 'rb') as file:
                 if embedding_model:
                     topic_model = joblib.load(file)
-                    topic_model.embedding_model = select_backend(embedding_model, topic_model.verbose)
+                    topic_model.embedding_model = select_backend(embedding_model, verbose=topic_model.verbose)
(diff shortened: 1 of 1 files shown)


MaartenGr:
> I've changed these to check whether the model (or any model in the case of merge_models) is set to be verbose. I did consider just always running them as non-verbose, but I thought that really if you're loading an existing model, BERTopic deciding to use a different model by itself is even worse than usual.

Agreed, keeping the changes you made makes sense here. 

> Of course this also indicates these code paths don't have test coverage as this would've failed otherwise. Not sure if that should also be addressed.

Yep, there is plenty that hasn't been added to the tests yet but mostly they are the different embedding/representation models. Some of these are tested but definitely not all.

MaartenGr:
Also, thanks for the fix!

PR #2060, opened 2024-06-21
Simplify zero-shot topic modeling

### LINKED ISSUE #1967: Issues with Zero-shot Topic Modeling regarding outliers and future operations
I am using zero-shot topic modeling and have run into some issues (a subset of which has already been brought up in passed issues), plus I have some concerns about correctness of the resulting topic model. I have two strategies in mind to resolve these, one of which I have already mostly written as a patch to the process for my project.

My flow is roughly:
 1. `topic_model = BERTopic(...)`
 2. `topic_model.fit(...)`
 3. `topic_model.reduce_topics(...)`
 4. `topic_model.reduce_outliers(...)`
 5. `topic_model.update_topics(...)`

## The Problems

Let's consider the case that some zero-shot topics have matched with some of the documents. The next step is to continue with clustering and normal topic modeling for the remaining documents in `self`.

In `_combine_zeroshot_topics`, a dummy model for the zero-shot topics is made and fitted to generate topic representations. This is merged with `self` (which current represents the topic model for un-assigned topics that went on to be clustered) using `BERTopic.merge_models` to create `merged_model`. Some cleanup is done before attributes in `self` are replaced with those in `merged_model`.

The final topics are 0 through `$#_matching_zeroshot_topics_-_one` and `$#_matching_zeroshot_topics` through `$#_matching_zeroshot_topics_+_#_clusters_-_1`. The outlier topic is now treated as a normal topic with ID `$#_matching_zeroshot_topics`.

### Outliers

The first issue is the loss of the outlier topic. The `merged_model._outliers`  is `0`, which makes its way into `self._outliers`.
This attribute is now wrong and it will no longer correctly be used in future operations as an offset to exclude the outlier topic.
The `-1` topic is now gone, so we have lost any meaningful reference to outliers existing. Future operations look at topic `-1` for identifying outlier documents. This issue is outlined in https://github.com/MaartenGr/BERTopic/pull/1957.

If we were to call `reduce_outliers` on a model without outliers, there is no validation. I have not looked into other strategies, but for the `c-tf-idf` strategy, the function will fail on `bow_doc = self.vectorizer_model.transform(outlier_docs)` because `outlier_docs` is empty. A subset of the `reduce_outliers` issue is described in https://github.com/MaartenGr/BERTopic/issues/1771.

### `self.c_tf_idf_`

Any future operation that references the `self.c_tf_idf_` matrix no longer works, because `self.c_tf_idf_ == None`, which is a carry-over from the `merged_model`. In my workflow, the `self.c_tf_idf_` matrix will be set correctly if I `reduce_topics`, because the final step is to extract the topic representations, which includes fitting `self.vetcorizer_model` and `self.ctfidf_model` and setting `self.c_tf_idf_`.

### Configuration-related attributes

We have now see outliers and the C-TF-IDF matrix not persist through the merge. Any other configuration-related attributes that may be important later are also not persisted. In my case, the relevant ones are `self.zeroshot_topic_list` , `self.vetcorizer_model`, and `self.ctfidf_model`, but there are many more. These get reset to their defaults. This especially becomes a problem if representations are ever updated again (which they are a couple more times in my workflow), because important properties like `reduce_frequent_words=True` in the `ClassTfidfTransformer` are no longer set.

This is a direct consequence of
```
self.__dict__.clear()
self.__dict__.update(merged_model.__dict__)
```
in `_combine_zeroshot_topics`.

### Correctness of representations

The representations in the `merged_model`, which a copied over into `self` are effective largely a concatenation of the represenations from the `zeroshot_model` and `self` (i.e. the clustered model). These representations are calculated in part with the `c_tf_idf_` matrix, which is produced in part from the size of the vocabulary. The problem is the vocabulary in `zeroshot_model` and `self` are only subsets of the real vocabulary, which includes *all* documents. As a user, I expect my zero-shot topics to be used at the very start as a sort of filtering step but mostly not have any effects later on (except the zero-shot topic labels persisting). This implies I expect representations to be calculated the way they would normally be, which is to use *all* documents.

While the representations calculated in each model separately are not necessarily wrong, I would argue it would be more correct for the represenations to be calculated considering *all* documents. That is, I could call `self._extract_topics(...)` after fitting, and the resulting representations (barring the special handling of zero-shot topic labels) should not change.

### `topic_embeddings_`

For this one, I am not 100% sure yet whether it is an issue. I would expect that upon merging the `zeroshot_model` and `self`, the `merged_model.topic_embeddings_` should essentially just be a concatenation of the two sets of topic embeddings (plus maybe some quirks relating to where that outlier topic goes). In my debugging, it appears that the topic embeddings from `zeroshot_model.topic_embeddings_` do not change, but the ones for the topics from clustering (from `self.topic_embeddings_`) do change. And that the cosine similarity of those with the ones in the new `merged_model` is not even high. Perhaps I am missing something when doing this analysis, or perhaps `BERTopic.merge_models` truly is not preserving topic embeddings, which I believe is wrong, as topic embeddings are **only** a function of the documents in each topic (and their embeddings) in this case. Maybe it's related to the `mappings` parameter in `_create_topic_vectors`? Still debugging this one...

### `reduce_outliers`

I discussed a lack of validation for the existence of outliers, leading to unclear errors. This function should start with a validation check to look for `-1 in self.topics_` (or really any attribute that has a mapping from topic IDs), *or* a more explicit solution is to check the value of `self._outliers`.

This private attribute can fall out of sync with reality, so I propose deriving it from one of the other attributes. This can look like
```
@property
def _outliers() -> int:
    # alternative solution which is less efficient:
    # return int(-1 in self.topics_)
    # other mapping attributes like self.topic_sizes_ can be used
    return int(-1 in self.topic_labels_)  # O(1) operation
```

One of the specific reasons I propose this is that in my flow, I calculate new topic IDs for all of my documents using `topic_model.reduce_outliers(...)`. This does not change the model at all, so I set `topic_model.topics_ = topic_model.reduce_outliers(...)`. I will later call `topic_model.update_topics(...)` to update the representations (which also sets the `self._outliers` attribute, but if I look for whether the `topic_model` still has outliers between these operations, `topic_model._outliers` is no longer a reliable source of information.

###  `_combine_zeroshot_topics`

This is where the bulk of my fixes through patching have been for the previously described problems.

#### Persisting of attributes from `self`

I described why I believe confguration-related attributes should not get cleared out. Obviously, (the majority of) attributes that were affected by fitting should be overriden in `self` by those in `merged_model`.  To be more specific,
```
# Public attributes
self.topics_ = None
self.probabilities_ = None
self.topic_sizes_ = None
self.topic_mapper_ = None
self.topic_representations_ = None
self.topic_embeddings_ = None
self.topic_labels_ = None
self.custom_labels_ = None
self.c_tf_idf_ = None
self.representative_images_ = None
self.representative_docs_ = {}
self.topic_aspects_ = {}
```
are the majority of attributes  that *should* get overridden. I have also added `umap_model` and `hdbscan_model` as attributes that should come from `merged_model`, but I am less sure about the implications of these.

I copy all attributes that are *not* these into a dictionary. The end of the function is changed to 
```
# Update the class internally
self.__dict__.clear()
self.__dict__.update(merged_model.__dict__)
####### Patched Changes #######
self.__dict__.update(properties_to_keep)
###############################
```
to keep those configuration-related properties. The `vectorizer_model` and `ctf_idf` model are kept for their configurations, even though they were only fitted to the clustered documents. This could be problematic. The result of this patch includes `self._outliers` being set properly.

#### Missing Outlier Topic

I have described that the resulting model has the zero-shot topics starting at ID `0` followed by the clustered topic, including the outlier topic. Right after the merge happens, the outlier topic is actually in the correct spot with ID `-1` but a gap exists after the zero-shot topics for where it will be moved to. I let this and related operations to the representation-related attributes run.

If `self._outliers` is set to 1, I will manually move the outlier topic back to ID `-1`. This involves three operations. For each of 
```
[
    merged_model.topic_labels_,
    merged_model.topic_sizes_,
    merged_model.representative_docs_,
    merged_model.topic_representations_
]
```
the a new mapping from `-1` to the info of the outlier topic (which is at ID `$#_matching_zeroshot_topics` ) is made. The mappings at  `$#_matching_zeroshot_topics_+_1` (all of the non-outlier topics that came from clustering) are shifted by 1 to a lower topic number. After the shift, the last mapping (ID `$#_matching_zeroshot_topics_+_#_clusters_-_1`) is deleted.

The same concept is implemented for `merged_model.topics_` (topic assignment for each document) but through the use of `numpy` operations before it is converted back to a `list`.

For an implementation that I think would be more appropriate to be merged into this repo, I think the `BERTopic.merge_models` should be changed to get this correct from the start.

#### Missing `c_tf_idf_` Matrix and Lack of Correctness

Both of these issues can be handled in the same operation, which can be summarized by `merged_model._extract_topics(all_docs, all_embeddings)`.

We want to update the representations in the merged model, which fits the `vectorizer_model` and `ctfidf_model` and sets the `c_tf_idf_` matrix, but this time with *all documents*. The `_combine_zeroshot_topics` function only accepts the assigned documents (`assigned_documents`), assigned embeddings (referred to in this function as `embeddings`), and remaining documents (`documents`). To get all embeddings, I patch the `fit_transform` function to pass in the embeddings of the clustered documents. In `_combine_zeroshot_topics`, `embeddings` is moved to `assigned_embeddings` to keep the naming consistent with `_zeroshot_topic_modeling`, and `embeddings` is now used for the embeddings of clustered documents (`documents`).

Before extracting topics, I set 
```
merged_model.ctfidf_model = self.ctfidf_model
merged_model.vectorizer_model = self.vectorizer_model
```
so we use the right configuration for those.

I also make sure afterwards to `merged_model.topic_labels_.update(zeroshot_topic_labels)` so they do not get overridden by the topic words. I have set `zeroshot_topic_labels` as a mapping from topic ID to label **only** for the zeroshot-topics.

I have not tested it yet, but I should also `_save_representative_docs` to update the representative docs, as these are a function of the `c_tf_idf_` matrix and the `vectorizer_model` and `ctfidf_model` models that are different now (by using *all* docs) than when representative docs were originally calculated (when the matrix and models were from vocabulary subsets in the zeroshot_model and self (i.e. the clustered model).

### `fit_transform`

So that seems like changes to `_combine_zeroshot_topics` are lengthy, potentially incomplete and computationally inefficient (representation mode(s) are run again and the representations from `zeroshot_model` and `self` before merge are essentially thrown out. A simpler solution that I have not written or detailed extensively is to avoid this merging of models in the first place.

The idea starts with separating assigned documents and assigned embeddings from the document and embeddings to be clustered -- same as the current implementation. The next steps are dimensionality reduction and clustering -- no changes yet.

Now, instead of performing representation-related operations on just the topics that came from clustering, we can merge the topics from clustering with the topics from zero-shot topic modeling. Some care needs to be taken to ensure the outlier topic stays at -1, but I do not think it matters much if zero-shot topics come next or get tacked onto the end (in fact, this alternate ordering in proposed in https://github.com/MaartenGr/BERTopic/issues/1771). The representations will be calculated using *all* topics and *all* documents. There would be no need for `_combine_zeroshot_topics`, but it perhaps could be repurposed for the combining of zero-shot topics with the topics from clusters after the clustering step. The `vectorizer_model` and `ctfidf_model` will be fitted to *all* documents, and `self.c_tf_idf_` will include the total vocabulary.

Now, operations that reference these (`reduce_topics` and `reduce_outliers` in my case) will (mostly) work, making no distinction between zero-shot topics and topics from clustering. There will still need to be some extra care whenever topic representations are updated (like in `reduce_topics`) to not override the zero-shot topic labels.

Couple more extra bit of care I can think of needs to be taken:
- When fitting, `self.nr_topics` may be set. I believe from the perspective of the user, this should apply to *all* topics, not just ones from clustering, as it currently the case. There should be validation that `self.nr_topics` exceeds the number of zero-shot topics that matched. If this is not the case, the user should be prompted to either raise `self.nr_topics` or raise the `zeroshot_min_similarity`.
- The returned probabilities and related `self._map_probabilities(probabilities, original_topics=True)`  needs to have the right shape. I have not dealt with these, but I fear these could already have problems in the current implementation when zero-shot topic modeling is used, because the probabilities appear to be only for the clustered documents, not all documents.

## Final Discussion

- Are there reasons for the quirks I have described in the problems section that are not explained in the referenced issue and PR?
- Do you see flaws in my description of my current patches?
- Does my alternative implementation of zero-shot topic modeling make sense, or are there specific reasons for why it is implemented as the merging of two `BERTopic` topic models?
- Anything else relevant to know or discuss?

ianrandman:
* Fix downstream operations after zero-shot topic modeling
  * No longer perform merging of models for zero-shot. The zero-shot topics act as a prefiltering step before clustering. The zero-shot topics are combined with the clustered topics immediately after clustering.
  * The C-TF-IDF model is fitted on all documents now, regardless of whether they belong to clustered or zero-shot topics.
* Validate number of topics when using zero-shot topic modeling
* Remove check for `type(self.hdbscan_model) != BaseCluster` when checking whether model is zero-shot
  * The HDBSCAN model is replaced with a `BaseCluster` during `fit_transform()` during zero-shot topic modeling.
* Derive `self._outliers` rather than tracking it to maintain alignment using `@property`
* Derive zero-shot labels when requested rather than tracking it using `@property`
* Fix typos related to `topic_to`, `topics_from` for mapping
* Validate existence of outliers in `reduce_outliers()`
* Maintain zero-shot topics while reducing topics
  * If original topics contain one or more zero-shot topics, the new topic keeps the best zero-shot topic if the cosine similarity with the topic meets `zeroshot_min_similarity`. Otherwise, the calculated representation is used.

Fixes #1967


bertopic/_bertopic.py
```diff
@@ -256,22 +256,41 @@ def __init__(self,
         self.topic_mapper_ = None
         self.topic_representations_ = None
         self.topic_embeddings_ = None
-        self.topic_labels_ = None
+        self.topic_id_to_zeroshot_topic_idx = {}
```
MaartenGr, line 259: I'm using the `_` at the end of the variable to distinguish between parameters/public attributes/private attributes. Could you update this so it is inline with the structure of the public attributes? Unless you believe this variable should not be accessed by the user at all, then we can convert it to a private attribute.

bertopic/_bertopic.py
```diff
@@ -439,14 +466,23 @@ def fit_transform(self,
...
+        if self._is_zeroshot() and len(assigned_documents) > 0:
+            self.hdbscan_model = BaseCluster()
+            sim_matrix = cosine_similarity(embeddings, np.array(self.topic_embeddings_))
+
+            if self.calculate_probabilities:
+                probabilities = sim_matrix
+            else:
+                # Use `topics_before_reduction` because `self.topics_` may have already been updated from
+                # reducing topics, and the original probabilities are needed for `self._map_probabilities()`
+                probabilities = sim_matrix[np.arange(len(documents)), topics_before_reduction]
```
MaartenGr, line 480: Could this be simplified by moving it to the end of the function and running something like:

```python
            self.hdbscan_model = BaseCluster()
            _, probabilities = self.transform(assigned_documents, embeddings)
```

That way, it would actually be a more proper `.fit` + `.transform` combo as you typically see in scikit-learn and reduces the lines of code a bit.

bertopic/_bertopic.py
```diff
@@ -3565,168 +3593,62 @@ def _zeroshot_topic_modeling(self, documents: pd.DataFrame, embeddings: np.ndarr
...
-        * Only clustered topics were found which will only return the clustered topic model
-        * Both zero-shot and clustered topics were found which will return a merged model
-          * This merged model is created using the `merge_models` function which will ignore
-            the underlying UMAP and HDBSCAN models
-
-        Arguments:
-            documents: Dataframe with documents and their corresponding IDs
-            assigned_documents: Dataframe with documents and their corresponding IDs
-                                that were assigned to a zero-shot topic
-            embeddings: The document embeddings
```
MaartenGr, line 3608: I would like to keep the docstrings here as it helps me (and others) develop. We could also move this to `._zeroshot_topic_modeling` and add some small docstrings here to indicate that this is only run when we have both zeroshot topics and non-zeroshot topics.

MaartenGr (commented):
Thank you for your work on this! I left a couple of small comments. Other than that, can you run `ruff`? With a PR that was recently merged, we now use `ruff` for the formatting/linting.

bertopic/_bertopic.py
ianrandman: This goes back to our discussion on using cosine similarities for probabilities after zero-shot topic modeling.

> They are indeed not perfectly aligned with one another. With flat probabilities/similarities, this is not an issue as we can simply give back the similarity of the assigned topic. With topic-doc probabilities/similarities, I'm not sure how to solve that issue without attempting to create a separate model (i.e., logistic regression) that is trained (and perhaps even overfitted) on the embeddings (X) and assigned topic (y) to extract hopefully more accurate probabilities.

If using `.transform`, the assignments, and the corresponding probabilities will not be aligned with the current state of assignments, because the current state came from a mix of zero-shot and clustering, and `self.topic_embeddings_` was not used at all when making these assignments.

We agreed this misalignment is fine. We will return the cosine similarities for the current assignments, even if calculating similarities using `self.topic_embeddings_` would show different assignments. I am not sure it is possible to reuse `.transform` for this purpose. I'm open to suggestions, but having that bit of duplicated code seems like the simplest solution.

On a related note, I do not think my solution accounts for outliers.
I have
```
probabilities = sim_matrix[
                    np.arange(len(documents)), topics_before_reduction
                ]
```
but I think it should be
```
probabilities = sim_matrix[
                    np.arange(len(documents)), np.array(topics_before_reduction) + self._outliers
                ]
```
Does this make sense to you?

bertopic/_bertopic.py
ianrandman: You're right. Forgot about this. Made it private.

bertopic/_bertopic.py
ianrandman: I had removed the docstring when making big changes and forgotten to write a new one. Added.

ianrandman:
> Thank you for your work on this! I left a couple of small comments. Other than that, can you run `ruff`? With a PR that was recently merged, we now use `ruff` for the formatting/linting.

Thanks for pointing out the recent incorporation with `ruff`. I'll have to start using that in my own projects.

I have fixed my changes with `ruff`, fixed the merge confict, and resolved a couple of your comments. Please mark them resolved if my changes look good. There is only the remaining discussion about probabilities.

freddyheppell:
It looks like this is still failing because I think you only ran one of the two Ruff commands.

Ruff has `format`, which replaces e.g. Black and `check` which replaces e.g. flake8. If you run `ruff check --fix` it should autofix the remaining issues and let you know which need manual fixing. You can also run `make format` and `make lint` as shortcuts.

bertopic/_bertopic.py
MaartenGr: > From a user perspective, I would expect the assignments from .fit_transform to be identical to the assignments from .fit followed by .transform.

Agreed, that wasn't the case previously (also due to some specific behavior of HDBSCAN) but that would definitely be preferred. 

> I'm open to suggestions, but having that bit of duplicated code seems like the simplest solution. 

Ah right, I totally missed that! Yep, let's keep it as is. 

> On a related note, I do not think my solution accounts for outliers.

Yeah, that should indeed be `+ self._outliers` I believe since the lowest potential value is -1 there which does nasty things when indexing.

MaartenGr:
I believe the code check in python 3.8 failed because its not familiar with the `tuple[..., ....]` type hints. I believe replacing them with either `from typing import Tuple` should work or simply removing those type hints.

MaartenGr:
@ianrandman Awesome, everything passed and I think we addressed all the comments we had. Just to be 100% sure, shall I go ahead and merge this?

ianrandman:
> @ianrandman Awesome, everything passed and I think we addressed all the comments we had. Just to be 100% sure, shall I go ahead and merge this?

Yes, all good to merge if it looks good to you. Happy to be done with this :).

MaartenGr:
@ianrandman Awesome, thank you for taking the time the last couple of works to work on this. It is greatly appreciated and hopefully this will also make it easier for you to use BERTopic instead of your own fork. If there are any other changes you would like to see, please let me know!

PR #2065, opened 2024-06-23
Fixes to chain strategy example in outlier_reduction.md

reuning:
# What does this PR do?

The chain strategies example in the outlier reduction has `new_topics` being acted on in the first step and then `topics` acted on in the second step. I'm pretty sure this is supposed to be other way around. This fixes that. 

## Before submitting
- [ X] This PR fixes a typo or improves the docs (if yes, ignore all other checks!).


MaartenGr:
Yep, you are completely right! Thank you for the PR 😄

PR #2068, opened 2024-06-25
Add git blame ignore

freddyheppell:
Adds git blame ignore to prevent most code in the repo being attributed to #2033.

GitHub will pick up on this by default.

Using Git locally, you can use this by passing the `--ignore-revs-file` argument e.g.
```sh
$ git blame bertopic/_bertopic.py --ignore-revs-file .git-blame-ignore-revs
```
or run this config to use by default
```sh
$ git config blame.ignoreRevsFile .git-blame-ignore-revs
```


MaartenGr:
Awesome, thanks for the PR!

PR #2085, opened 2024-07-15
Add missing pass of `self.verbose` when selecting backend

### LINKED ISSUE #2079: Verbose setting gets lost when embeddings and embedding model provided

### Desribe the bug

If providing both embeddings and an embedding model during `BERTopic` construction, the verbosity of `logger` will be set to `WARNING` during `.fit` regardless of the `verbose` parameter.

https://github.com/MaartenGr/BERTopic/blob/bf1fedd5bb14bbf282d5051e849d2b06abaa8b51/bertopic/_bertopic.py#L427-L440

On the last line, `self.verbose` is not passed like it is when `embeddings` is not passed. Patching the function to include this parameter works as expected. I can make a PR with this fix, unless this was a purposeful design decision.

To be clear, my proposed solution is to change line 440 from
```
self.embedding_model = select_backend(self.embedding_model, language=self.language) 
```
to
```
self.embedding_model = select_backend(self.embedding_model, language=self.language, verbose=self.verbose) 
```

I think the strategy of modifying the level of the top-level logger based on the `verbose` setting within a `BERTopic` instance may be error-prone and subject to race condition problems if multiple threads are working with `BERTopic` instances with varying `verbose` settings. I believe a better long term solution to logging is constructing a new logger per `BERTopic` instantiation, but that is out of the scope of this issue.

### BERTopic Version

0.16.2

ianrandman:
Fixes #2079

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -437,7 +437,9 @@ class BERTopic:
             logger.info("Embedding - Completed \u2713")
         else:
             if self.embedding_model is not None:
-                self.embedding_model = select_backend(self.embedding_model, language=self.language)
+                self.embedding_model = select_backend(
+                    self.embedding_model, language=self.language, verbose=self.verbose
+                )
 
         # Guided Topic Modeling
         if self.seed_topic_list is not None and self.embedding_model is not None:


MaartenGr:
Thanks for the PR!

PR #2105, opened 2024-08-01
Fix #2102

### LINKED ISSUE #2102: incorrect result by topic_model.get_topic_info()   due to zeroshot_topic_list was set

### Desribe the bug

We have df with 5 clusters
[df.csv](https://github.com/user-attachments/files/16426710/df.csv)

and zeroshot_topic_list with 3 clusters
[zeroshot_topic_list.csv](https://github.com/user-attachments/files/16426709/zeroshot_topic_list.csv)

There are 150 holidays, 100 work and 50 code messages in the dataset. The final result in topic_model.get_topic_info() Name does not match Representation because Name is passed in the sort in which it is set in zeroshot_topic_list (1 code, 2 work, 3 holidays), and Representation is sorted by the number of messages in the cluster from larger to smaller (3 holidays, 2 work, 1 code).

### Reproduction

```
import pandas as pd
import os
from bertopic import BERTopic
from umap import UMAP


root_dir = os.path.abspath(os.path.join(os.path.dirname(__file__)))

representation_model = KeyBERTInspired()

sentence_model = SentenceTransformer(
    model_name_or_path=f"{root_dir}/models/paraphrase-multilingual-mpnet-base-v2"
) 

zeroshot_topic_list = pd.read_csv(
    f"{root_dir}/zeroshot_topic_list.csv",
    sep=";",
    index_col=0,
)

df = pd.read_csv(
    f"{root_dir}/df.csv",
    sep=";",
    index_col=0,
)

zeroshot_topic_list = list(
    zeroshot_topic_list["Name"].str.replace(r"^\d+_", "", regex=True)
)

embeddings = sentence_model.encode(list(df["text"]), show_progress_bar=False)
umap_model = UMAP(
    n_neighbors=15,
    n_components=5,
    min_dist=0.0,
    metric="cosine",
    random_state=42,
)
topic_model = BERTopic(
    language="multilingual",
    umap_model=umap_model,
    embedding_model=sentence_model,
    top_n_words=20,
    n_gram_range=(1, 2),
    min_topic_size=50,
    nr_topics=None,  
    vectorizer_model=None,
    representation_model=representation_model,
    zeroshot_min_similarity=0.7,
)
topics, probs = topic_model.fit_transform(list(df["text"]), embeddings)
print(topic_model.get_topic_info())

```

### BERTopic Version

0.16.3

MaartenGr:
# What does this PR do?

Fixes #2102

This fixes a number of things:
* Incorrect ordering of the topic embeddings by using "Old_ID" instead of "ID"
* Unneeded `ValueError` when `nr_topics="auto"` is used combined with zero-shot topic modeling
* Attempting to transform UMAP even when no documents where clustered (only zero-shot topics)
* Add `TopicMapper` at the right moment so that **all** topics are added to the mapper and not only a subset

I updated the way probabilities were returned as I faced some issues with selecting the correct topics. I might change it back after more testing. 

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -313,7 +313,7 @@ class BERTopic:
         if self._is_zeroshot():
             # Need to correct labels from zero-shot topics
             topic_id_to_zeroshot_label = {
-                topic_id: self.zeroshot_topic_list[zeroshot_topic_idx]
+                self.topic_mapper_.get_mappings()[topic_id]: self.zeroshot_topic_list[zeroshot_topic_idx]
                 for topic_id, zeroshot_topic_idx in self._topic_id_to_zeroshot_topic_idx.items()
             }
             topic_labels.update(topic_id_to_zeroshot_label)
@@ -453,10 +453,12 @@ class BERTopic:
             documents, embeddings, assigned_documents, assigned_embeddings = self._zeroshot_topic_modeling(
                 documents, embeddings
             )
+            
             # Filter UMAP embeddings to only non-assigned embeddings to be used for clustering
-            umap_embeddings = self.umap_model.transform(embeddings)
+            if len(documents) > 0:
+                umap_embeddings = self.umap_model.transform(embeddings)
 
-        if len(documents) > 0:  # No zero-shot topics matched
+        if len(documents) > 0:
             # Cluster reduced embeddings
             documents, probabilities = self._cluster_embeddings(umap_embeddings, documents, y=y)
             if self._is_zeroshot() and len(assigned_documents) > 0:
@@ -467,7 +469,6 @@ class BERTopic:
             # All documents matches zero-shot topics
             documents = assigned_documents
             embeddings = assigned_embeddings
-        topics_before_reduction = self.topics_
 
         # Sort and Map Topic IDs by their frequency
         if not self.nr_topics:
@@ -505,17 +506,11 @@ class BERTopic:
             sim_matrix = cosine_similarity(embeddings, np.array(self.topic_embeddings_))
 
             if self.calculate_probabilities:
-                probabilities = sim_matrix
+                self.probabilities_ = sim_matrix
             else:
-                # Use `topics_before_reduction` because `self.topics_` may have already been updated from
-                # reducing topics, and the original probabilities are needed for `self._map_probabilities()`
-                probabilities = sim_matrix[
-                    np.arange(len(documents)),
-                    np.array(topics_before_reduction) + self._outliers,
-                ]
-
-        # Resulting output
-        self.probabilities_ = self._map_probabilities(probabilities, original_topics=True)
+                self.probabilities_ = np.max(sim_matrix, axis=1)
+        else:
+            self.probabilities_ = self._map_probabilities(probabilities, original_topics=True)
         predictions = documents.Topic.to_list()
 
         return predictions, self.probabilities_
@@ -3835,18 +3830,23 @@ class BERTopic:
 
         # Check that if a number of topics was specified, it exceeds the number of zeroshot topics matched
         num_zeroshot_topics = len(assigned_documents["Topic"].unique())
-        if self.nr_topics and not self.nr_topics > num_zeroshot_topics:
-            raise ValueError(
-                f"The set nr_topics ({self.nr_topics}) must exceed the number of matched zero-shot topics "
-                f"({num_zeroshot_topics}). Consider raising nr_topics or raising the "
-                f"zeroshot_min_similarity ({self.zeroshot_min_similarity})."
-            )
+        if self.nr_topics != "auto":
+            if self.nr_topics and not self.nr_topics > num_zeroshot_topics:
+                raise ValueError(
+                    f"The set nr_topics ({self.nr_topics}) must exceed the number of matched zero-shot topics "
+                    f"({num_zeroshot_topics}). Consider raising nr_topics or raising the "
+                    f"zeroshot_min_similarity ({self.zeroshot_min_similarity})."
+                )
 
         # Select non-assigned topics to be clustered
         documents = documents.iloc[non_assigned_ids]
         documents["Old_ID"] = documents["ID"].copy()
         documents["ID"] = range(len(documents))
         embeddings = embeddings[non_assigned_ids]
+
+        if len(documents) == 0:
+            self.topics_ = assigned_documents["Topic"].values.tolist()
+            self.topic_mapper_ = TopicMapper(self.topics_)
 
         logger.info("Zeroshot Step 1 - Completed \u2713")
         return documents, embeddings, assigned_documents, assigned_embeddings
@@ -4108,7 +4108,10 @@ class BERTopic:
             topic_embeddings = []
             topics = documents.sort_values("Topic").Topic.unique()
             for topic in topics:
-                indices = documents.loc[documents.Topic == topic, "ID"].values
+                if self._is_zeroshot():
+                    indices = documents.loc[documents.Topic == topic, "Old_ID"].values
+                else:
+                    indices = documents.loc[documents.Topic == topic, "ID"].values
                 indices = [int(index) for index in indices]
                 topic_embedding = np.mean(embeddings[indices], axis=0)
                 topic_embeddings.append(topic_embedding)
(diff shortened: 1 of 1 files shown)


bertopic/_bertopic.py
```diff
@@ -4527,9 +4530,13 @@ def _sort_mappings_by_frequency(self, documents: pd.DataFrame) -> pd.DataFrame:
             documents: Updated dataframe with documents and the mapped
                        and re-ordered topic ids
         """
-        self._update_topic_size(documents)
+        # No need to sort if it's the first pass of zero-shot topic modeling
+        nr_zeroshot = len(self._topic_id_to_zeroshot_topic_idx)
+        if self._is_zeroshot and not self.nr_topics and nr_zeroshot > 0:
+            return documents
```
ianrandman, line 4536: I thought the plan was to update `self._topic_id_to_zeroshot_topic_idx` in `._sort_mappings_by_frequency` if zeroshot, and then *not* use `self.topic_mapper_.get_mappings()` in `.topic_labels_`?

bertopic/_bertopic.py
```diff
@@ -467,7 +469,6 @@ def fit_transform(
             # All documents matches zero-shot topics
             documents = assigned_documents
             embeddings = assigned_embeddings
-        topics_before_reduction = self.topics_
 
         # Sort and Map Topic IDs by their frequency
         if not self.nr_topics:
```
ianrandman, line 474: Why does sorting mappings *not* occur if `nr_topics` is passed?

bertopic/_bertopic.py
```diff
@@ -4108,7 +4108,10 @@ def _create_topic_vectors(
             topic_embeddings = []
             topics = documents.sort_values("Topic").Topic.unique()
             for topic in topics:
-                indices = documents.loc[documents.Topic == topic, "ID"].values
+                if self._is_zeroshot():
+                    indices = documents.loc[documents.Topic == topic, "Old_ID"].values
+                else:
+                    indices = documents.loc[documents.Topic == topic, "ID"].values
```
ianrandman, line 4114: I would suggest that instead of this check, we just correct `ID` when combining zeroshot topics after https://github.com/MaartenGr/BERTopic/blob/fa65bc4f92f76308ec116c55dafa8c9c661ab446/bertopic/_bertopic.py#L3909
I think `documents.ID = documents.Old_ID` would work. Any reason you can think of why this wouldn't be correct?

bertopic/_bertopic.py
MaartenGr: Ah right, I wasn't familiar enough with `self._topic_id_to_zeroshot_topic_idx` so wasn't sure the best way to change it. This seemed faster and is working now but if you have suggestions on how to change it, feel free to share.

bertopic/_bertopic.py
MaartenGr: The `topics_before_reduction` were used for calculating the probabilities and after I made a couple of changes, it started giving index errors when attempting to access the similarity matrix.

bertopic/_bertopic.py
MaartenGr: I tried it before here:

https://github.com/MaartenGr/BERTopic/blob/2353f4c21d74e33e34e30dbae938304bff094792/bertopic/_bertopic.py#L3847

But it didn't work there. Let me try your suggestion there and see if it helps.

bertopic/_bertopic.py
MaartenGr: Specifically, if I add `nr_topics="auto"` when training BERTopic using the latest commit in this PR, I get the following error:

```python

# KeyBERT
keybert_model = KeyBERTInspired()

# Pass the above models to be used in BERTopic
topic_model = UpdatedBERTopic(
    embedding_model=embedding_model,
    umap_model=umap_model,
    hdbscan_model=hdbscan_model,
    verbose=True,
    zeroshot_min_similarity=.5,
    nr_topics="auto",
    calculate_probabilities=True,
    representation_model=keybert_model
)
topic_model = topic_model.fit(abstracts, embeddings)
```

Then that gives me the following error:

```bash
IndexError                                Traceback (most recent call last)
Cell In[13], line 18
      6 # Pass the above models to be used in BERTopic
      7 topic_model = UpdatedBERTopic(
      8     embedding_model=embedding_model,
      9     umap_model=umap_model,
   (...)
     16     representation_model=keybert_model
     17 )
---> 18 topic_model = topic_model.fit(abstracts, embeddings)

File [~\Documents\Projects\BERTopic\bertopic\_bertopic.py:364](http://localhost:8888/lab/tree/Documents/Projects/~/Documents/Projects/BERTopic/bertopic/_bertopic.py#line=363), in BERTopic.fit(self, documents, embeddings, images, y)
    322 def fit(
    323     self,
    324     documents: List[str],
   (...)
    327     y: Union[List[int], np.ndarray] = None,
    328 ):
    329     """Fit the models (Bert, UMAP, and, HDBSCAN) on a collection of documents and generate topics.
    330 
    331     Arguments:
   (...)
    362     ```
    363     """
--> 364     self.fit_transform(documents=documents, embeddings=embeddings, y=y, images=images)
    365     return self

Cell In[12], line 300, in UpdatedBERTopic.fit_transform(self, documents, embeddings, images, y)
    292     else:
    293         # Use `topics_before_reduction` because `self.topics_` may have already been updated from
    294         # reducing topics, and the original probabilities are needed for `self._map_probabilities()`
    295         probabilities = sim_matrix[
    296             np.arange(len(documents)),
    297             np.array(topics_before_reduction) + self._outliers,
    298         ]
--> 300 self.probabilities_ = self._map_probabilities(probabilities, original_topics=True)
    301 predictions = documents.Topic.to_list()
    303 return predictions, self.probabilities_

File [~\Documents\Projects\BERTopic\bertopic\_bertopic.py:4581](http://localhost:8888/lab/tree/Documents/Projects/~/Documents/Projects/BERTopic/bertopic/_bertopic.py#line=4580), in _map_probabilities(self, probabilities, original_topics)
   4578             if to_topic != -1 and from_topic != -1:
   4579                 mapped_probabilities[:, to_topic] += probabilities[:, from_topic]
-> 4581         return mapped_probabilities
   4583 return probabilities

IndexError: index 71 is out of bounds for axis 1 with size 71

```

MaartenGr:
@ianrandman If you have the time, could you take a look at the open comments? If we can resolve them, I can start putting out a new minor version with the fix.

ianrandman:
> @ianrandman If you have the time, could you take a look at the open comments? If we can resolve them, I can start putting out a new minor version with the fix.

Sorry about the delay. I can get back to you within a day with more detailed comments. Some quick comments I can give now are that it looks like your solution uses my temp fix for the topic labels property rather than changing the zeroshot idx dict  as topic mapping change. Using the temp fix results in incorrectness (I think) with topic reduction because the implemtation there is to change the zeroshot idx dict as the topic mapping changed.

I think that anytime the zeroshot idx dict is looked at, it should be correct without needing to apply the mapping from the topic mapper. As far as I know, all that is needed to achieve that, given the current implementation, is mentioned in https://github.com/MaartenGr/BERTopic/pull/2105#discussion_r1701933665.

Let me know your thoughts on this and whether I (hopefully don't) need to provide more detail.

MaartenGr:
@ianrandman Thank you for the quick reply! 

> Sorry about the delay. I can get back to you within a day with more detailed comments. Some quick comments I can give now are that it looks like your solution uses my temp fix for the topic labels property rather than changing the zeroshot idx dict as topic mapping changes. Using the temp fix results in incorrectness (I think) with topic reduction because the implemtation there is to change the zeroshot idx dict as the topic mapping changed.

Hmm, I tested the topic reduction a couple of times and although it works it indeed does merge the zero-shot topics. 

> I think that anytime the zeroshot idx dict is looked at, it should be correct without needing to apply the mapping from the topic mapper. As far as I know, all that is needed to achieve that, given the current implementation, is mentioned in https://github.com/MaartenGr/BERTopic/pull/2105#discussion_r1701933665.

I've been looking through the zeroshot idx dict but I'm just not sure where and how to update it for your proposed fix. It seems that I do not have a sufficient enough grasp on the logic here to make the changes.

ianrandman (approved):

Issue #2123, opened 2024-08-17
Custom labels for get_topic_tree


MaartenGr:
### Discussed in https://github.com/MaartenGr/BERTopic/discussions/2122

<div type='discussions-op-text'>

<sup>Originally posted by **clfkenny** August 17, 2024</sup>
Hi there, I noticed the `visualize_hierarchy` method has a `custom_labels` parameter. Is there anything similar to this for the `get_topic_tree` method? I use the LLM generated labels as the custom labels so the hierarchy tree is much easier to interpret. Thanks! </div>

This seems like a good first issue for someone wanting to contribute to this repository. It has a very similar structure to the `visualize_hierarchy` functionality, so I believe it should be straightforward to implement. Unfortunately, I have a large backlog and cannot prioritize this, so if anyone wants to help out then that would be great!

PriyaDCosta:
May I take this up? I have used BERTopic a lot for my projects and I am happy to contribute.

MaartenGr:
@PriyaDCosta Thank you for helping out! I believe there is already an open PR at #2125 that is currently tackling this. If I'm not mistaken, a few small changes are needed over there before we can proceed.

PR #2125, opened 2024-08-19
Add custom labels to get topic tree

clfkenny:
# What does this PR do?

`get_topic_tree` function currently does not accept custom_labels as a parameter whereas the other visualization methods do. Therefore this change aims to fix that issue so custom labels can be accepted and make the visualization more intuitive. 

I really enjoyed looking through the codebase here. My first PR here so still learning the project structure, would appreciate any feedback. Thanks!

Fixes # https://github.com/MaartenGr/BERTopic/issues/2123

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -1852,7 +1852,7 @@ class BERTopic:
         hier_topics: pd.DataFrame,
         max_distance: float = None,
         tight_layout: bool = False,
-        custom_labels: Union[bool, str] = False
+        custom_labels: Union[bool, str] = False,
     ) -> str:
         """Extract the topic tree such that it can be printed.
 
@@ -1908,14 +1908,14 @@ class BERTopic:
         child_right_ids = hier_topics.Child_Right_ID.astype(int)
 
         # Get the new parent labels generated from `hierarchical_topics`
-        new_left_labels = {int(row['Child_Left_ID']): row['Child_Left_Name'] for idx, row in hier_topics.iterrows()}
-        new_right_labels = {int(row['Child_Right_ID']): row['Child_Right_Name'] for idx, row in hier_topics.iterrows()}
+        new_left_labels = {int(row["Child_Left_ID"]): row["Child_Left_Name"] for idx, row in hier_topics.iterrows()}
+        new_right_labels = {int(row["Child_Right_ID"]): row["Child_Right_Name"] for idx, row in hier_topics.iterrows()}
 
         if custom_labels:
             left_labels = {}
             if isinstance(custom_labels, str):
                 for topic, kws_info in self.topic_aspects_[custom_labels].items():
-                    label = '_'.join([kw[0] for kw in kws_info[:5]]) # displaying top 5 kws
+                    label = "_".join([kw[0] for kw in kws_info[:5]])  # displaying top 5 kws
                     left_labels[topic] = label
             elif self.custom_labels_ is not None and custom_labels:
                 left_labels = {topic_id: label for topic_id, label in enumerate(self.custom_labels_, -self._outliers)}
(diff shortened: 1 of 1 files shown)


bertopic/_bertopic.py
```diff
@@ -1897,9 +1903,40 @@ def get_topic_tree(
...
+        # Get the new parent labels generated from `hierarchical_topics`
+        new_left_labels = {int(row["Child_Left_ID"]): row["Child_Left_Name"] for idx, row in hier_topics.iterrows()}
+        new_right_labels = {int(row["Child_Right_ID"]): row["Child_Right_Name"] for idx, row in hier_topics.iterrows()}
+
+        if custom_labels:
+            left_labels = {}
+            if isinstance(custom_labels, str):
+                for topic, kws_info in self.topic_aspects_[custom_labels].items():
+                    label = "_".join([kw[0] for kw in kws_info[:5]])  # displaying top 5 kws
+                    left_labels[topic] = label
```
MaartenGr, line 1919: Should we add a check for when a user attempts to use a specific aspect but it is not found in `self.topic_aspects_`? I can imagine that a user (myself included) would mistype the name and the resulting error might not be entirely clear.

bertopic/_bertopic.py
```diff
@@ -1897,9 +1903,40 @@ def get_topic_tree(
...
+            elif self.custom_labels_ is not None and custom_labels:
+                left_labels = {topic_id: label for topic_id, label in enumerate(self.custom_labels_, -self._outliers)}
```
MaartenGr, line 1921: I just wanted to mention that I am going to steal that enumerate trick. That is so much nicer coding than how I have approached it thus far.

bertopic/_bertopic.py
```diff
@@ -1897,9 +1903,40 @@ def get_topic_tree(
...
+
+        # Get the new parent labels generated from `hierarchical_topics`
+        new_left_labels = {int(row["Child_Left_ID"]): row["Child_Left_Name"] for idx, row in hier_topics.iterrows()}
+        new_right_labels = {int(row["Child_Right_ID"]): row["Child_Right_Name"] for idx, row in hier_topics.iterrows()}
+
+        if custom_labels:
+            left_labels = {}
+            if isinstance(custom_labels, str):
+                for topic, kws_info in self.topic_aspects_[custom_labels].items():
+                    label = "_".join([kw[0] for kw in kws_info[:5]])  # displaying top 5 kws
```
MaartenGr, line 1918: Note that some topic aspects might not have 5 keywords but actually a single label. In those cases, the `kws_info[0]` is likely to be filled with a label (or even a summary) and each instance in `kws_info[1:]` will be an empty string. As a result, you might get the following label: "artificial intelligence____".

MaartenGr (commented):
Thanks for the PR! The code looks solid and I left minor comments.

One other thing, would you mind expanding the tests at `tests/test_variations/test_hierarchy` (`test_tree`) a bit to include the `custom_labels=True`?

bertopic/_bertopic.py
clfkenny: This makes sense, I can add a check here. One thing I noticed when looking through how custom labels were handled in the plotting functions is that they all follow a pretty similar pattern:

https://github.com/MaartenGr/BERTopic/blob/510c15e0f3f3c6481f21087d9461e5fb5e8b89af/bertopic/plotting/_topics.py#L67
https://github.com/MaartenGr/BERTopic/blob/510c15e0f3f3c6481f21087d9461e5fb5e8b89af/bertopic/plotting/_heatmap.py#L99
https://github.com/MaartenGr/BERTopic/blob/510c15e0f3f3c6481f21087d9461e5fb5e8b89af/bertopic/plotting/_datamap.py#L123
https://github.com/MaartenGr/BERTopic/blob/510c15e0f3f3c6481f21087d9461e5fb5e8b89af/bertopic/plotting/_distribution.py#L62
https://github.com/MaartenGr/BERTopic/blob/510c15e0f3f3c6481f21087d9461e5fb5e8b89af/bertopic/plotting/_documents.py#L139

Maybe a more generic and reusable function can be created for this, which would check if the user specified aspect exists in `self.topic_aspects_`

MaartenGr: You're right, it makes no sense to do that check here but not at all of those other instances. For now, I'm alright either way. I can imagine wanting to stick with what I already use everywhere else for consistency as this might be a bit out of the scope of this particular PR. So it's up to you.

Issue #1767, opened 2024-01-22
How to get more than 3 representative docs per topic via get_topic_info()?


youssefabdelm:
Hey, thank you so much for making this library! Super awesome. 

I've seen a bunch of issues here requesting this but haven't found a straightforward easy way to specify it via get_topic_info() as it contains a lot of the information I need. I wish there was a parameter in there like get_topic_info(number_of_representative_documents=3) that I could modify.

MaartenGr:
The documents themselves are not saved within BERTopic in part to reduce memory requirements, so it would not be possible to run something like `.get_topic_info(numberof_representative_documents=3)`. You are always using c-TF-IDF since it is part of the default pipeline, so `._extract_representative_docs` should work.

syGOAT:
> so `._extract_representative_docs` should work.

I couldn't find an example in BETopic doc. Could you please provide an example?

MaartenGr:
I think there is a nice example [here](https://github.com/MaartenGr/BERTopic/issues/1679). It might be nice to have an additional function that re-calculates the representative documents since this question seems to appear frequently.

PR #2130, opened 2024-08-28
Add nr_repr_docs parameter

robechun:
# What does this PR do?
Adds a parameter that you can pass around for getting the `nr_repr_docs`. This helps when attempting to get a higher # of representative docs for a given topic.

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -370,6 +370,7 @@ class BERTopic:
         embeddings: np.ndarray = None,
         images: List[str] = None,
         y: Union[List[int], np.ndarray] = None,
+        nr_repr_docs: int = None,
     ) -> Tuple[List[int], Union[np.ndarray, None]]:
         """Fit the models on a collection of documents, generate topics,
         and return the probabilities and topic per document.
@@ -381,6 +382,7 @@ class BERTopic:
             images: A list of paths to the images to fit on or the images themselves
             y: The target class for (semi)-supervised modeling. Use -1 if no class for a
                specific instance is specified.
+            nr_repr_docs: [optional] Number of representative docs to keep for each topic.
 
         Returns:
             predictions: Topic predictions for each documents
@@ -487,7 +489,7 @@ class BERTopic:
                 custom_documents = self._reduce_topics(custom_documents)
 
             # Save the top 3 most representative documents per topic
-            self._save_representative_docs(custom_documents)
+            self._save_representative_docs(custom_documents, nr_repr_docs=nr_repr_docs)
         else:
             # Extract topics by calculating c-TF-IDF
             self._extract_topics(documents, embeddings=embeddings, verbose=self.verbose)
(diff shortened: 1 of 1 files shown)


MaartenGr:
Thank you for this pull request! I believe there were a number of issues open discussing this, which issue does this PR refer to? I remember there was some discussion on how to approach this there. That perspective will help me understand what was needed and what decisions we made before going through this PR.

robechun:
I didn't actually look at past issues—it was more from my own usage that I found this added parameter to be helpful. That said, it looks like [this issue](https://github.com/MaartenGr/BERTopic/issues/1767) is the most pertinent

MaartenGr:
Ah, then you might have missed the [contributing guideline](https://github.com/MaartenGr/BERTopic/blob/master/CONTRIBUTING.md)! I think it's first important to establish what exactly is needed. 

From the issue you shared as well as the connected one, it seems that users also want to re-calculate the number of representative documents *after* training the model.

Also, what made you choose to add this parameter to `.transform` rather than the init? This seems like a parameter that should go there.

robechun:
Whoops, missed that part on creating an issue! 😅 

WRT what is exactly needed—AFAICT, I think there are three needs/wants based on the specific usage pattern.

1. Where users simply use `fit_transform` to get the topic clusters (in my use-case)
2. Where users want to get representative docs for online topic modeling
3. Where users want to specify `nr_repr_docs` when they call `merge_topics`

This PR would solve for 1 and 3 (given the `nr_repr_docs` can now be added as an input parameter to either `fit_transform` and `merge_topics`) but not necessarily 2.

As for why I included it in `transform` vs `init`—my mental model was the following:
* `init` is used for instantiating the different "lego" pieces for the BERTopic orchestration (such as clustering algorithm, representation model, etc).
* `fit_transform` or `merge_models` is where the user can provide specifics about what they want from the model—for example, the model itself remains the same, but I could run the same model with different documents to produce different outputs. Similarly, I could run the same model with the same documents and different `nr_repr_docs` to extract a slightly different output (namely the set of representative docs).

Let me know if you think I'm off in any way

MaartenGr:
> This PR would solve for 1 and 3 (given the nr_repr_docs can now be added as an input parameter to either fit_transform and merge_topics) but not necessarily 2.

I think it would indeed be nice to give users the ability to choose the number of representative documents. Wouldn't it be easier to update the number of representative documents *after* fitting the model? That way, it would solve all 3 problems without the need to actually needing to refit the model constantly. The representative documents should not affect the representations (except for LLM-based representation models which has separate parameters for this) so it does not feel like it should be a parameter related to either init or transform. It might be easiest and more related to the task itself to put it *after* training the model. 

> As for why I included it in transform vs init—my mental model was the following:

Generally, we would put these kinds of parameter in the `init` since it is part of the parameter space of the model and is not specific to the behavior of `transform`. However, as mentioned above, it might actually be better to go with a new function that allows you to update the number of representative documents after having created the model.

MaartenGr:
@kpetrianakis Not yet in its current state as is shown in my response above. I believe a new function might be the most API friendly as it allows you to easily update the number of representative **after** training. So there is no new training pipeline needed, just extraction.

PR #2145, opened 2024-09-12
added system prompt for openai

Leo-LiHao:
# What does this PR do?
Add system prompt for OpenAI Representation model, see Issue #2146 

## Before submitting

### DIFF
--- a/bertopic/representation/_cohere.py
+++ b/bertopic/representation/_cohere.py
@@ -33,7 +33,9 @@ Topic:
 Sample texts from this topic:
 [DOCUMENTS]
 Keywords: [KEYWORDS]
-Topic name:"""
+Provide the topic name directly without any explanation."""
+
+DEFAULT_SYSTEM_PROMPT = "You are designated as an assistant that identify and extract high-level topics from texts."
 
 
 class Cohere(BaseRepresentation):
@@ -51,6 +53,8 @@ class Cohere(BaseRepresentation):
                 NOTE: Use `"[KEYWORDS]"` and `"[DOCUMENTS]"` in the prompt
                 to decide where the keywords and documents need to be
                 inserted.
+        system_prompt: The system prompt to be used in the model. If no system prompt is given,
+                       `self.default_system_prompt_` is used instead.
         delay_in_seconds: The delay in seconds between consecutive prompts
                                 in order to prevent RateLimitErrors.
         nr_docs: The number of documents to pass to OpenAI if a prompt
@@ -107,8 +111,9 @@ class Cohere(BaseRepresentation):
     def __init__(
         self,
         client,
-        model: str = "xlarge",
+        model: str = "command-r",
         prompt: str = None,
+        system_prompt: str = None,
         delay_in_seconds: float = None,
         nr_docs: int = 4,
         diversity: float = None,
@@ -118,7 +123,9 @@ class Cohere(BaseRepresentation):
         self.client = client
         self.model = model
         self.prompt = prompt if prompt is not None else DEFAULT_PROMPT
+        self.system_prompt = system_prompt if system_prompt is not None else DEFAULT_SYSTEM_PROMPT
         self.default_prompt_ = DEFAULT_PROMPT
+        self.default_system_prompt_ = DEFAULT_SYSTEM_PROMPT
         self.delay_in_seconds = delay_in_seconds
         self.nr_docs = nr_docs
         self.diversity = diversity
@@ -160,14 +167,14 @@ class Cohere(BaseRepresentation):
             if self.delay_in_seconds:
                 time.sleep(self.delay_in_seconds)
 
-            request = self.client.generate(
+            request = self.client.chat(
                 model=self.model,
-                prompt=prompt,
+                preamble=self.system_prompt,
+                message=prompt,
                 max_tokens=50,
-                num_generations=1,
                 stop_sequences=["\n"],
             )
-            label = request.generations[0].text.strip()
+            label = request.text.strip().replace("Topic name: ", "")
             updated_topics[topic] = [(label, 1)] + [("", 0) for _ in range(9)]
 
         return updated_topics
--- a/bertopic/representation/_cohere.py
+++ b/bertopic/representation/_cohere.py
@@ -33,9 +33,9 @@ Topic:
 Sample texts from this topic:
 [DOCUMENTS]
 Keywords: [KEYWORDS]
-Provide the topic name directly without any explanation."""
+Topic name:"""
 
-DEFAULT_SYSTEM_PROMPT = "You are designated as an assistant that identify and extract high-level topics from texts."
+DEFAULT_SYSTEM_PROMPT = "You are an assistant that extracts high-level topics from texts."
 
 
 class Cohere(BaseRepresentation):
@@ -174,7 +174,7 @@ class Cohere(BaseRepresentation):
                 max_tokens=50,
                 stop_sequences=["\n"],
             )
-            label = request.text.strip().replace("Topic name: ", "")
+            label = request.text.strip()
             updated_topics[topic] = [(label, 1)] + [("", 0) for _ in range(9)]
 
         return updated_topics
--- a/bertopic/representation/_llamacpp.py
+++ b/bertopic/representation/_llamacpp.py
@@ -8,13 +8,34 @@ from bertopic.representation._utils import truncate_document
 
 
 DEFAULT_PROMPT = """
-Q: I have a topic that contains the following documents:
+This is a list of texts where each collection of texts describe a topic. After each collection of texts, the name of the topic they represent is mentioned as a short-highly-descriptive title
+
+Example 1:
+Sample texts from this topic:
+- Traditional diets in most cultures were primarily plant-based with a little meat on top, but with the rise of industrial style meat production and factory farming, meat has become a staple food.
+- Meat, but especially beef, is the word food in terms of emissions.
+- Eating meat doesn't make you a bad person, not eating meat doesn't make you a good one.
+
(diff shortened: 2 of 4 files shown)


bertopic/representation/_cohere.py
```diff
@@ -33,7 +33,9 @@
 Sample texts from this topic:
 [DOCUMENTS]
 Keywords: [KEYWORDS]
-Topic name:"""
+Provide the topic name directly without any explanation."""
```
MaartenGr, line 36: Why did you make this change? In my experience, by providing the model with a pre-fix, there is no need for mentioning that it should provide the topic name without any explanation.

MaartenGr (commented):
Thanks for updating these two models! I left a small comment. 

Other than that, I believe we will need to look at the following:
* [HuggingFace](https://github.com/MaartenGr/BERTopic/blob/master/bertopic/representation/_textgeneration.py)
* [LangChain](https://github.com/MaartenGr/BERTopic/blob/master/bertopic/representation/_langchain.py)
* [Llama-cpp-python](https://github.com/MaartenGr/BERTopic/blob/master/bertopic/representation/_llamacpp.py)

With HuggingFace, we might need to change the formatting such that we use messages instead:

```python
messages = [
    {
        "role": "system",
        "content": MY_SYSTEM_PROMPT,
    },
    {"role": "user", "content": prompt},
 ]
```

For LangChain, I'm okay with skipping this for now since it uses a more complex structure and the API here might need to be updated (see https://python.langchain.com/v0.1/docs/modules/model_io/chat/quick_start/#messages-in---message-out).

For llama-cpp-python, it seems straightforward and we should use [`.create_chat_completion`](https://github.com/abetlen/llama-cpp-python?tab=readme-ov-file#chat-completion) instead of directly calling the model.

bertopic/representation/_cohere.py
Leo-LiHao: Based on my experiments, the pre-fix prompt works well for completion models but not with (small) chat models. You can try different prompts locally to see the difference and feel free to change/iterate the prompt.

Leo-LiHao:
@MaartenGr I updated the code for llama-cpp

bertopic/representation/_cohere.py
MaartenGr: Let's keep it the same it was. Updating the prompt would require another PR with additional testing. Also, this changes the way the few-shot prompt was created.

bertopic/representation/_llamacpp.py
```diff
@@ -8,13 +8,34 @@
...
+- It took a month longer to deliver than was advised...
+
+Keywords: deliver weeks product shipping long delivery received arrived arrive week
+Topic name: Shipping and delivery issues
+---
+Extract the topic name from the following documents:
+Sample texts from this topic:
 [DOCUMENTS]
+Keywords: [KEYWORDS]
+Provide the extracted topic name directly without any explanation."""
```
MaartenGr, line 36: Let's structure this the same way as was originally done in `_cohere.py`

bertopic/representation/_cohere.py
```diff
@@ -33,7 +33,9 @@
...
+
+DEFAULT_SYSTEM_PROMPT = "You are designated as an assistant that identify and extract high-level topics from texts."
```
MaartenGr, line 38:
```suggestion
DEFAULT_SYSTEM_PROMPT = "You are an assistant that extracts high-level topics from texts."
```

bertopic/representation/_llamacpp.py
```diff
@@ -8,13 +8,34 @@
...
 
-The topic is described by the following keywords: '[KEYWORDS]'.
-
-Based on the above information, can you give a short label of the topic?
-A: """
+DEFAULT_SYSTEM_PROMPT = "You are designated as an assistant that identify and extract high-level topics from texts."
```
MaartenGr, line 38:
```suggestion
DEFAULT_SYSTEM_PROMPT = "You are an assistant that extracts high-level topics from texts."
```

bertopic/representation/_llamacpp.py
```diff
@@ -93,14 +114,15 @@ def __init__(
         self,
         model: Union[str, Llama],
         prompt: str = None,
+        system_prompt: str = None,
         pipeline_kwargs: Mapping[str, Any] = {},
         nr_docs: int = 4,
         diversity: float = None,
-        doc_length: int = None,
+        doc_length: int = 100,
```
MaartenGr, line 121: Let's keep the value as it was, otherwise we have to do this for all instances of `doc_length` across all models.

bertopic/representation/_llamacpp.py
```diff
@@ -93,14 +114,15 @@ def __init__(
...
         tokenizer: Union[str, Callable] = None,
     ):
         if isinstance(model, str):
-            self.model = Llama(model_path=model, n_gpu_layers=-1, stop="Q:")
+            self.model = Llama(model_path=model, n_gpu_layers=-1, stop="\n", chat_format="llama-2")
```
MaartenGr, line 125: What makes you choose `llama-2` as the default chat template? Should we choose one at all or are that perhaps more common templates like `ChatML`?

bertopic/representation/_llamacpp.py
```diff
@@ -110,6 +132,7 @@ def __init__(
                 "local LLM or a ` llama_cpp.Llama` object."
             )
         self.prompt = prompt if prompt is not None else DEFAULT_PROMPT
+        self.system_prompt = system_prompt if system_prompt is not None else DEFAULT_SYSTEM_PROMPT
```
MaartenGr, line 135: Should there also be a `self.default_system_prompt_`?

bertopic/representation/_llamacpp.py
```diff
@@ -150,33 +173,31 @@ def extract_topics(
...
+            topic_description = self.model.create_chat_completion(
+                  messages=[
+                      {"role": "system", "content": self.system_prompt},
+                      {
+                          "role": "user",
+                          "content": prompt
+                      }
+                  ], ** self.pipeline_kwargs
+            )
+            label = topic_description["choices"][0]["message"]["content"].strip().replace("Topic name: ", "")
```
MaartenGr, line 186: I think we can remove `.replace("Topic name: ", "")` if the prompt is updated (see comment above).

bertopic/representation/_llamacpp.py
```diff
@@ -150,33 +173,31 @@ def extract_topics(
...
-                    to_replace += f"- {doc}\n"
-                prompt = prompt.replace("[DOCUMENTS]", to_replace)
+        prompt = self.prompt
+        if "[KEYWORDS]" in prompt:
+            prompt = prompt.replace("[KEYWORDS]", keywords)
+        if "[DOCUMENTS]" in prompt:
+            to_replace = ""
+            for doc in docs:
+                to_replace += f"- {doc}\n"
+            prompt = prompt.replace("[DOCUMENTS]", to_replace)
```
MaartenGr, line 201: If we update this piece of code here, shouldn't we also do this for the other models, like `_cohere.py`?

bertopic/representation/_openai.py
```diff
@@ -48,6 +48,8 @@
 topic: <topic label>
 """
 
+DEFAULT_SYSTEM_PROMPT = "You are designated as an assistant that identify and extract high-level topics from texts."
```
MaartenGr, line 51:
```suggestion
DEFAULT_SYSTEM_PROMPT = "You are an assistant that extracts high-level topics from texts."
```

bertopic/representation/_utils.py
```diff
@@ -44,8 +44,8 @@ def decode(self, doc_chunks):
         truncated_document: A truncated document
     """
     if doc_length is not None:
-        if tokenizer == "char":
-            truncated_document = document[:doc_length]
+        if tokenizer == "char" or tokenizer is None:
+            truncated_document = document[:doc_length] + " (...)"
```
MaartenGr, line 48: Could you explain why this was added? I believe something like this was addressed in https://github.com/MaartenGr/BERTopic/pull/2181

MaartenGr (commented):
Apologies for the late reply, I was off for the last couple of weeks. 

I left some comments here and there to make sure we are on the same page for all changes.

bertopic/representation/_llamacpp.py
Leo-LiHao: I followed the readme in llamacpp, I have changed it to `ChatML`

bertopic/representation/_llamacpp.py
Leo-LiHao: I think it is a bug in the old implementation which does not replace `[DOCUMENTS]` for `DEFAULT_PROMPT`. I have replaced `_create_prompt` in `_llamacpp.py` as the one in `_cohere.py`

bertopic/representation/_utils.py
Leo-LiHao: Yes, some issue which was addressed in #2181

Leo-LiHao:
> Apologies for the late reply, I was off for the last couple of weeks.
> 
> I left some comments here and there to make sure we are on the same page for all changes.

@MaartenGr Thanks for your comments, I incorporated them and updated the code.

MaartenGr:
@Leo-LiHao Thanks! I noticed a couple of changes to creating the prompt that we might want in all other models but it's alright to keep that as is for now. At some point, I might make a utility function for creating the prompt that is used across all models to make sure everything remains stable. 

For now, let's merge!

MaartenGr:
Ah, I see the linting has an error. Could you check that?

Leo-LiHao:
@MaartenGr Thank you! I fixed the lint error and it is read for merge now.

MaartenGr:
Awesome, thanks for taking the time! I just merged it 🥳

PR #2180, opened 2024-10-14
Fixed issue #2144

### LINKED ISSUE #2144: Representations from representation_models are generated twice when nr_topics="auto".
### Discussed in https://github.com/MaartenGr/BERTopic/discussions/2142

<div type='discussions-op-text'>

<sup>Originally posted by **siddharthtumre** September 10, 2024</sup>
Hi @MaartenGr 

While using nr_topics="auto", I can see that the representations from representation_models are calculated after reducing the number of topics. This increases the run time as a LLM is prompted twice i.e., before and after reducing topics.

I believe generating representations after reducing the topics should be sufficient.</div>

That's indeed what is happening currently. I believe it should be straightforward to implement that so this might be a good first issue for those wanting to help out.

For now, what you could also do is run the model first without any representation models and then use .update_topics without your selected representation model (e.g, LLM) to create the representations. Then the LLM would run only once and you get the same results.

pipa666:
# What does this PR do?

Fixes #2144 (issue) by optimizing the topic extraction process when using `fit_transform()` with `nr_topics="auto"` or `int` for reducing topics. The main improvements are:

1. Avoiding double calculation of topic representations, especially when using LLM-based representations.
2. Introducing a `calculate_representation` parameter in the `_extract_topics` method to control when representations are calculated.
3. Modifying the `fit_transform` method to calculate representations only after topic reduction (if needed).

These changes significantly improve performance, particularly when using computationally expensive representation models like LLMs.

Additionally, this PR addresses an edge case where `self.nr_topics >= initial_nr_topics`. In this scenario, the current implementation extracts topics but doesn't sort the mappings by frequency. This results in a list of topics with unsorted indexes. While topic information is still displayed sorted by frequencies, the topic indexes may not match their frequency rank. 

This is easily solvable by adding `self._sort_mappings_by_frequency(documents)` in  `def _reduce_topics()` in line 4369. But runs into problems when `pytest tests\test_representation\test_representations.py` tests for the edge case `self.nr_topics >= initial_nr_topics`
Thus, some fix in how this edge cases are tested is required.

## Before submitting
    - Extra log messages were added to improve debugging
    - Test is successful as it is now. However, they do not account for the issue described above

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -478,21 +478,20 @@ class BERTopic:
         if documents.Document.values[0] is None:
             custom_documents = self._images_to_text(documents, embeddings)
 
-            # Extract topics by calculating c-TF-IDF
-            self._extract_topics(custom_documents, embeddings=embeddings)
-            self._create_topic_vectors(documents=documents, embeddings=embeddings)
-
-            # Reduce topics
+            # Extract topics by calculating c-TF-IDF, reduce topics if needed, and get representations.
+            self._extract_topics(custom_documents, embeddings=embeddings, calculate_representation=not self.nr_topics)
             if self.nr_topics:
                 custom_documents = self._reduce_topics(custom_documents)
+            self._create_topic_vectors(documents=documents, embeddings=embeddings)
 
             # Save the top 3 most representative documents per topic
             self._save_representative_docs(custom_documents)
-        else:
-            # Extract topics by calculating c-TF-IDF
-            self._extract_topics(documents, embeddings=embeddings, verbose=self.verbose)
 
-            # Reduce topics
+        else:
+            # Extract topics by calculating c-TF-IDF, reduce topics if needed, and get representations.
+            self._extract_topics(
+                documents, embeddings=embeddings, verbose=self.verbose, calculate_representation=not self.nr_topics
+            )
             if self.nr_topics:
                 documents = self._reduce_topics(documents)
 
@@ -3972,6 +3971,7 @@ class BERTopic:
         embeddings: np.ndarray = None,
         mappings=None,
         verbose: bool = False,
+        calculate_representation: bool = True,
     ):
         """Extract topics from the clusters using a class-based TF-IDF.
 
@@ -3980,18 +3980,29 @@ class BERTopic:
             embeddings: The document embeddings
             mappings: The mappings from topic to word
             verbose: Whether to log the process of extracting topics
+            calculate_representation: Whether to extract the topic representations
 
         Returns:
             c_tf_idf: The resulting matrix giving a value (importance score) for each word per topic
         """
         if verbose:
-            logger.info("Representation - Extracting topics from clusters using representation models.")
+            action = "Representation" if calculate_representation else "Topics"
+            logger.info(
+                f"{action} - Extracting topics from clusters{' using representation models' if calculate_representation else ''}."
+            )
+
         documents_per_topic = documents.groupby(["Topic"], as_index=False).agg({"Document": " ".join})
         self.c_tf_idf_, words = self._c_tf_idf(documents_per_topic)
-        self.topic_representations_ = self._extract_words_per_topic(words, documents)
+        self.topic_representations_ = self._extract_words_per_topic(
+            words,
+            documents,
+            calculate_representation=calculate_representation,
+            calculate_aspects=calculate_representation,
+        )
         self._create_topic_vectors(documents=documents, embeddings=embeddings, mappings=mappings)
+
         if verbose:
-            logger.info("Representation - Completed \u2713")
+            logger.info(f"{action} - Completed \u2713")
 
     def _save_representative_docs(self, documents: pd.DataFrame):
         """Save the 3 most representative docs per topic.
@@ -4245,6 +4256,7 @@ class BERTopic:
         words: List[str],
         documents: pd.DataFrame,
         c_tf_idf: csr_matrix = None,
+        calculate_representation: bool = True,
         calculate_aspects: bool = True,
     ) -> Mapping[str, List[Tuple[str, float]]]:
         """Based on tf_idf scores per topic, extract the top n words per topic.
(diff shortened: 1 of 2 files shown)


pipa666:
Ok, so I solved the issue with the `test_representations.py::test_topic_reduction_edge_cases` 

It just needed one line of code to `old_documents = topic_model._sort_mappings_by_frequency(old_documents)`
Therefore, the test will actually check for ordered, and unchanged, old/new topics

I hope is enough and self-explanatory. I forced pushed into the same commit to keep git history clean

bertopic/_bertopic.py
```diff
@@ -3980,18 +3980,29 @@ def _extract_topics(
...
         Returns:
             c_tf_idf: The resulting matrix giving a value (importance score) for each word per topic
         """
         if verbose:
-            logger.info("Representation - Extracting topics from clusters using representation models.")
+            action = "Representation" if calculate_representation else "Topics"
+            logger.info(
+                f"{action} - Extracting topics from clusters{' using representation models' if calculate_representation else ''}."
+            )
+
```
MaartenGr, line 3993: Just to be sure I understood this correctly, are the two options as follows:

> "Representation - Extracting topics from clusters using representation models"
 
Does this refer to the case where **all** representations are calculated (and not just c-TF-IDF)?

> "Topics - Extracting topics from clusters"

Does this then refer to the case where **only** c-TF-IDF is calculate?

Also, perhaps it would be nice to create a variable for the second if/else statement the same way you did for the first. 

```python 
method = ' using representation models' if calculate_representation else ''
logger.info(f"{action} - Extracting topics from clusters{method}."
```

bertopic/_bertopic.py
```diff
@@ -3980,18 +3980,29 @@ def _extract_topics(
             embeddings: The document embeddings
             mappings: The mappings from topic to word
             verbose: Whether to log the process of extracting topics
+            calculate_representation: Whether to extract the topic representations
```
MaartenGr, line 3983: Shouldn't this be the "additional" topic representations instead since c-TF-IDF is already calculated? I'm struggling a bit with having calculated c-TF-IDF before or not. Now it seems like c-TF-IDF is not calculated at all.

bertopic/_bertopic.py
```diff
@@ -3980,18 +3980,29 @@ def _extract_topics(
...
+            )
+
         documents_per_topic = documents.groupby(["Topic"], as_index=False).agg({"Document": " ".join})
         self.c_tf_idf_, words = self._c_tf_idf(documents_per_topic)
-        self.topic_representations_ = self._extract_words_per_topic(words, documents)
+        self.topic_representations_ = self._extract_words_per_topic(
+            words,
+            documents,
+            calculate_representation=calculate_representation,
+            calculate_aspects=calculate_representation,
```
MaartenGr, line 4000: Just to be sure I understand this correctly. You added this here because when you only calculate the base representation  (`calculate_representation=False`) you do not want the additional representations (`calculate_aspects=False`), right?

If so, we might need to think about the term `calculate_representation` because it should refer to fine-tuning representations instead right? Or did you name it did way because technically, it might also be the default representation when you set it as `"Main"`?

bertopic/_bertopic.py
```diff
@@ -4468,7 +4488,7 @@ def _auto_reduce_topics(self, documents: pd.DataFrame, use_ctfidf: bool = False)
         # Update documents and topics
         self.topic_mapper_.add_mappings(mapped_topics, topic_model=self)
         documents = self._sort_mappings_by_frequency(documents)
-        self._extract_topics(documents, mappings=mappings)
+        self._extract_topics(documents, mappings=mappings, verbose=self.verbose)
```
MaartenGr, line 4491: I see you added verbosity here. I purposefully left this to False to prevent too much logging but was wondering why you think it was necessary to add it.

MaartenGr (commented):
Awesome PR! Very clean and focused on the most important components here. I left a couple of minor questions that have mostly to do with variable naming. 

One last question is whether that edge case still appears that we talked about previously. Was it fully prevented by adding what you showed here; 

```python 
            else:
                logger.info(
                    f"Topic reduction - Number of topics ({self.nr_topics}) is equal or higher than the clustered topics({len(self.get_topics())})."
                )
                documents = self._sort_mappings_by_frequency(documents)
                self._extract_topics(documents, verbose=self.verbose)
                return documents
```

pipa666:
> Awesome PR! Very clean and focused on the most important components here. I left a couple of minor questions that have mostly to do with variable naming.
> 
> One last question is whether that edge case still appears that we talked about previously. Was it fully prevented by adding what you showed here;
> 
> ```python
>             else:
>                 logger.info(
>                     f"Topic reduction - Number of topics ({self.nr_topics}) is equal or higher than the clustered topics({len(self.get_topics())})."
>                 )
>                 documents = self._sort_mappings_by_frequency(documents)
>                 self._extract_topics(documents, verbose=self.verbose)
>                 return documents
> ```

Yes,  that solves the edge case. The issue I had before was that test_representation.py would raise a false negative error, but i got that fixed by editing the test

bertopic/_bertopic.py
pipa666: Thats right.

Indeed is about fine tuning. But yea, the nomenclature here is confusing for me as well.

It could be changed to "fine_tune_representations"

I will gather my thoughts on one single comment bellow, since this touches on the other comments.

bertopic/_bertopic.py
pipa666: Indeed, the `"Topics - Extracting topics from clusters"` refers to the case where only the c-TF-IDF is calculated.
And I would suggest changing it to `"Extracting base representation from bag of words"`

I can make the change suggested, but please check my comment [below](https://github.com/MaartenGr/BERTopic/pull/2180#issuecomment-2425958936).

bertopic/_bertopic.py
pipa666: Yes, it is. 

The c-TF-IDF is always calculated, since is not conditional to that boolean argument. 
Is calculated in line 3995 `self.c_tf_idf_, words = self._c_tf_idf(documents_per_topic)` and then used to calculate the default representation in `def _extract_words_per_topic`
`if not self.representation_model or not calculate_representation:
            # Default representation: c_tf_idf + top_n_words
            topics = {label: values[: self.top_n_words] for label, values in topics.items()}`

pipa666:
Ok I closed this PR because i found a much cleaner solution. I will make a new one, you will notice it has a minimal change that might not even require major review. I will open the logging discussion as a new issue, to distinguish the two issues and make my point better.

Issue #1977, opened 2024-05-07
Should raise an Exception when tokenizer is not defined


timo-obrecht:
In **bertopic/representation/_utils.py**, line 57, `tokenizer` is possibly `None`. In this case, an exception asking the user to explicitly set tokenizer should be raised.

```python
    if doc_length is not None:
        if tokenizer == "char":
            truncated_document = document[:doc_length]
        elif tokenizer == "whitespace":
            truncated_document = " ".join(document.split()[:doc_length])
        elif tokenizer == "vectorizer":
            tokenizer = topic_model.vectorizer_model.build_tokenizer()
            truncated_document = " ".join(tokenizer(document)[:doc_length])
        elif hasattr(tokenizer, 'encode') and hasattr(tokenizer, 'decode'):
            encoded_document = tokenizer.encode(document)
            truncated_document = tokenizer.decode(encoded_document[:doc_length])
        return truncated_document
    return document
```

MaartenGr:
Thanks for this! It would indeed be much nicer to raise an expectation and explain what the user should do. If you want, a PR would be much appreciated.

PR #2181, opened 2024-10-14
Fixed Issue: #1977

SSivakumar12:
First of all, I want to take the opportunity to thank the awesome work being done by Martin and all contributors both past and present. This is a really good tool that we have used at work and has provided us with a lot of meaningful insights so I figured that I could do my bit to contribute to a tool that has helped me previously :)

This is my first PR and I sincerely hope it is not my last  especially within this repo since I really want to give back to this project. 

# What does this PR do?
Address the issue raised in issues #1977 which I believe has not been resolved yet.

Fixes # (issue)
I have added an edge case so that if a tokenizer isn't any of the expected patterns a ValueError is raised and produced a constructive error message to hopefully to prevent a repeat of the error. I did not update the documentation and add any tests. That being said, I am more than happy to add tests if deemed appropriate/valuable

## Before submitting
      - didn't update documentation but added a error message
      - no but happy to do so in a separate PR since I think it doesn't fall into the scope of this current PR?

### DIFF
--- a/bertopic/representation/_cohere.py
+++ b/bertopic/representation/_cohere.py
@@ -124,9 +124,10 @@ class Cohere(BaseRepresentation):
         self.diversity = diversity
         self.doc_length = doc_length
         self.tokenizer = tokenizer
+        validate_truncate_document_parameters(self.tokenizer, self.doc_length)
+
         self.prompts_ = []
 
-        _ = validate_truncate_document_parameters(self.tokenizer, self.doc_length)
 
     def extract_topics(
         self,
--- a/bertopic/representation/_cohere.py
+++ b/bertopic/representation/_cohere.py
@@ -128,7 +128,6 @@ class Cohere(BaseRepresentation):
 
         self.prompts_ = []
 
-
     def extract_topics(
         self,
         topic_model,
--- a/bertopic/representation/_cohere.py
+++ b/bertopic/representation/_cohere.py
@@ -126,6 +126,13 @@ class Cohere(BaseRepresentation):
         self.tokenizer = tokenizer
         self.prompts_ = []
 
+        if self.tokenizer is None and self.doc_length is not None:
+            raise ValueError(
+                "Please select from one of the valid options for the `tokenizer` parameter: \n"
+                "{'char', 'whitespace', 'vectorizer'} \n"
+                "If `tokenizer` is of type callable ensure it has methods to encode and decode a document \n"
+            )
+
     def extract_topics(
         self,
         topic_model,
--- a/bertopic/representation/_cohere.py
+++ b/bertopic/representation/_cohere.py
@@ -4,7 +4,7 @@ from tqdm import tqdm
 from scipy.sparse import csr_matrix
 from typing import Mapping, List, Tuple, Union, Callable
 from bertopic.representation._base import BaseRepresentation
-from bertopic.representation._utils import truncate_document
+from bertopic.representation._utils import truncate_document, validate_truncate_document_parameters
 
 
 DEFAULT_PROMPT = """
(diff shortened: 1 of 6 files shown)


MaartenGr:
Thank you for the PR! Is this exception raised whenever the documents are actually truncated or when you instantiate the related LLM? If it is the former, you would get an error after the clustering has been done which seems like wasted resources for the user. Would it perhaps be an idea to raise this error earlier?

SSivakumar12:
Good point! In my case right now, the error would be raised when I instantiate the relevant LLM. I will look to try and find a suitable place to raise the earlier and amend accordingly.

SSivakumar12:
Looking at an example implementation for an representation class such as [llamacpp](https://github.com/MaartenGr/BERTopic/blob/master/bertopic/representation/_llamacpp.py#L148),  the purpose of `truncate_document`, is to prepare the prompt so that it could be sent to the LLM?

I might be wrong so I do apologise but wouldn't this mean you wouldn't get an error after clustering? Therefore where it is reasonable?

bertopic/representation/_cohere.py
```diff
@@ -126,12 +126,7 @@ def __init__(
         self.tokenizer = tokenizer
         self.prompts_ = []
 
-        if self.tokenizer is None and self.doc_length is not None:
-            raise ValueError(
-                "Please select from one of the valid options for the `tokenizer` parameter: \n"
-                "{'char', 'whitespace', 'vectorizer'} \n"
-                "If `tokenizer` is of type callable ensure it has methods to encode and decode a document \n"
-            )
+        _ = validate_truncate_document_parameters(self.tokenizer, self.doc_length)
```
MaartenGr, line 129: Are you returning the `_` for a specific reason? Looking at the `validate_truncate_document_parameters`, there doesn't seem anything that is returned.

bertopic/representation/_utils.py
```diff
@@ -58,6 +58,15 @@ def decode(self, doc_chunks):
...
     return document
 
+def validate_truncate_document_parameters(tokenizer, doc_length) -> Union[None, ValueError]:
+    """validates parameters that are used in the function `truncate_document`"""
+    if tokenizer is None and doc_length is not None:
+        raise ValueError(
+            "Please select from one of the valid options for the `tokenizer` parameter: \n"
+            "{'char', 'whitespace', 'vectorizer'} \n"
+            "If `tokenizer` is of type callable ensure it has methods to encode and decode a document \n"
+        )
```
MaartenGr, line 68: Should we also include a check for the opposite? That someone does use a tokenizer but not a doc_length?

bertopic/representation/_utils.py
```diff
@@ -57,6 +58,21 @@ def decode(self, doc_chunks):
...
+    elif tokenizer is not None and doc_length is None:
+        raise ValueError(
+            "If `tokenizer` is provided, `doc_length` of type int must be provided as well."
+        )
+    else:
+        pass
```
MaartenGr, line 74: I believe this isn't necessary right?

SSivakumar12: No it is not necessary but I included it to be verbose in terms of the possible edgecase. That being said, happy to remove it.

MaartenGr: Let's remove it and I'll make sure to merge it after that! Other than that, it looks great!

MaartenGr:
@SSivakumar12 Could you check the linting? It is failing there.

SSivakumar12:
Apologies,  for the oversight, have resolved the linting issues locally and the build should pass on linting now

MaartenGr:
Awesome, everything is looking good. Thank you for your work on this!

PR #2213, opened 2024-11-13
Add LiteLLM as a representation model

MaartenGr:
# What does this PR do?

Adds [LiteLLM](https://docs.litellm.ai/) as a representation model, which should open up many LLMs to use.

To use this, you will need to install the openai package first:

`pip install litellm`

Then, get yourself an API key of any provider (for instance OpenAI) and use it as follows:

```python
import os
from bertopic.representation import LiteLLM
from bertopic import BERTopic

# set ENV variables
os.environ["OPENAI_API_KEY"] = "your-openai-key"

# Create your representation model
representation_model = LiteLLM(model="gpt-3.5-turbo")

# Use the representation model in BERTopic on top of the default pipeline
```

## Before submitting


bertopic/representation/_litellm.py
```diff
@@ -0,0 +1,176 @@
...
+        nr_docs: The number of documents to pass to LiteLLM if a prompt
+                 with the `["DOCUMENTS"]` tag is used.
+        diversity: The diversity of documents to pass to LiteLLM.
+                   Accepts values between 0 and 1. A higher
+                   values results in passing more diverse documents
+                   whereas lower values passes more similar documents.
+
+    Usage:
+
+    To use this, you will need to install the openai package first:
```
PipaFlores, line 49: Shouldn't it say 'install the **litellm** package first' ?

MaartenGr: It should! Thanks for catching this.

Skar0:
Hi! I’m just asking out of curiosity—would it make sense to use the LangChain connection here together with the LangChain representation instead of having a separate implementation altogether? I understand it might not be the preferred approach, but I came across [this](https://python.langchain.com/docs/integrations/chat/litellm/) and thought it might be worth mentioning :slightly_smiling_face:

MaartenGr:
@Skar0 Thanks for sharing! Although that is a perfectly reasonable approach, LangChain changes its API quite often and has a large set of dependencies to take into account. In contrast, LiteLLM is light in its dependencies and its API shouldn't change much since it adheres to OpenAIs offering. 

I figured that a lighter alternative is welcome in this case.

PR #2254, opened 2024-12-23
Update _topics.py

### LINKED ISSUE #2247: Issue Warning when plotting MDS with two few topics
### Feature request

I propose having a nice explainable warning that informs the user one cannot plot MDS with insufficient number of topics.

### Motivation

The issue was opened many times: #1914, #1010, #725, #378. This means people get confused by the original error. Having an integrated error issued after a check on the number of topics in the topic model would be beneficial.

### Your contribution

I looked into the code, but honestly didn't know where such a warning would be most appropriate and how to format it.

ajdapretnar:
Proposed solution to not having enough topics and the call visualize_topics() resulting in an Error.

# What does this PR do?

Fixes #2247.
Not sure whether this requires updating documentation or a test. If yes, happy to write it.

## Before submitting
- [X] Did you read the [contributor guideline](https://github.com/MaartenGr/BERTopic/blob/master/CONTRIBUTING.md)?
- [X] Was this discussed/approved via a Github issue? Please add a link to it if that's the case. #2247

### DIFF
--- a/bertopic/plotting/_topics.py
+++ b/bertopic/plotting/_topics.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pandas as pd
+import warnings
 from umap import UMAP
 from typing import List, Union
 from sklearn.preprocessing import MinMaxScaler
@@ -56,6 +57,9 @@ def visualize_topics(
     freq_df = freq_df.loc[freq_df.Topic != -1, :]
     if topics is not None:
         topics = list(topics)
+    elif len(topics) < 5:
+        warnings.warn("Too few topics to display.")
+        return
     elif top_n_topics is not None:
         topics = sorted(freq_df.Topic.to_list()[:top_n_topics])
     else:


bertopic/plotting/_topics.py
```diff
@@ -56,6 +57,9 @@ def visualize_topics(
     freq_df = freq_df.loc[freq_df.Topic != -1, :]
     if topics is not None:
         topics = list(topics)
+    elif len(topics) < 5:
+        warnings.warn("Too few topics to display.")
+        return
```
MaartenGr, line 62: Couple of questions:

* What makes you choose the value of < 5 here? Is it working for 4 topics but it is for 5?
* Should we raise an error rather than return nothing after a warning? By returning None, we might also have to change the typing. 
* Could you extend the description a bit more? Perhaps explain that we need more topics in order to properly visualize them?

MaartenGr (commented):
Thank you for the PR! I left a couple of small questions/comments.

PR #2280, opened 2025-02-05
fix: handling OpenAI returning None objects

### LINKED ISSUE #2176: OpenAI representation fails to produce output when response content is None

### Desribe the bug

I ran into issues when using the OpenAI representation as it sometimes produces a content of None, which then produced an error when trying to run:
label = response.choices[0].message.content.strip().replace("topic: ", "")

Which makes sense, since the content is not a string. 
I'm unable to generate a minimal example since this is due to the output of OpenAI GPT.

I see two ways to work around this, but both have their own downsides/impact on the results, maybe anyone else sees better option:

1. set the content to type string before processing it any further. With the major downside that the label will then be set to the string 'None'
2. use a try and except to extract the content, strip this and replace the 'topic:' part of the string. If this fails the label is set to a fixed value like an empty string (and producing a warning that his has happened)

For now I fixed it by creating an inherited customOpenAI representation class within my script where I used the second option as a solution. 

### Reproduction

```python
from bertopic import BERTopic

```

### BERTopic Version

0.16.4

jeaninejuliettes:
# What does this PR do?

Fixes #2176
OpenAI sometimes returns a None type object instead of a response object. This could happen when you, for instance, run into a rate limit issue or a content violation error. To prevent the whole run to fail, a check is added if the response object is not of type None, if the object is of type None, the label for that specific topic is set to "No label returned". This approach of handling an issue when generating the label for a topic was chosen in a previous iteration and is kept as is.

No documentation has been added with this PR since this is a very minor fix that doesn't influence documentation. There has been a comment added to the code in line with a previously iteration done to the code.
No tests have been added.

## Before submitting

### DIFF
--- a/bertopic/representation/_openai.py
+++ b/bertopic/representation/_openai.py
@@ -245,7 +245,8 @@ class OpenAI(BaseRepresentation):
 
                 # Check whether content was actually generated
                 # Addresses #1570 for potential issues with OpenAI's content filter
-                if hasattr(response.choices[0].message, "content"):
+                # Addresses #2176 for potential issues when openAI returns a None type object
+                if response and hasattr(response.choices[0].message, "content"):
                     label = response.choices[0].message.content.strip().replace("topic: ", "")
                 else:
                     label = "No label returned"


MaartenGr:
Thanks for the PR!

This is great, seems like a simple enough and straightforward solution for the fix without needing to overly document this.

PR #2306, opened 2025-03-14
Lightweight installation: use safetensors without torch

### LINKED ISSUE #2304: Lightweight installation: use safetensors without torch
### Feature request

Remove dependency on `torch` when loading the topic model saved with safetensors. 

### Motivation

I was happy to find [the guide](https://maartengr.github.io/BERTopic/getting_started/tips_and_tricks/tips_and_tricks.html#lightweight-installation) to lightweight bertopic installation (without `torch`), however, BERTopic.load() seems to depend on `torch` through safetensors. 

Removing torch from requirements gives 2x speedup to the build of docker container for me, since I am hosting the embedding model on another service, so I would really like this to be avoided if possible.

Specifically, the problem is [here](https://github.com/MaartenGr/BERTopic/blob/master/bertopic/_save_utils.py#L514) in _save_utils.py:

``` python
def load_safetensors(path):
    """Load safetensors and check whether it is installed."""
    try:
        import safetensors.torch  # <----
        import safetensors

        return safetensors.torch.load_file(path, device="cpu")
    except ImportError:
        raise ValueError("`pip install safetensors` to load .safetensors")
```

### Your contribution

I suggest using the function `safetensors.safe_open()` with `framework='numpy'` instead. This way the load() does not have the unnecessary requirement for torch to be installed.

hedgeho:
# What does this PR do?

This PR removes unnecessary imports of torch from model load/save code to enable lightweight installation (without torch) as described [here in docs](https://maartengr.github.io/BERTopic/getting_started/tips_and_tricks/tips_and_tricks.html#lightweight-installation).

It achieves this mainly by replacing calls to `safetensors.torch.[save|load]_file()` with `safetensors.numpy.[save|load]_file()`, and removing other calls to torch methods. 

Fixes #2304

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -4856,7 +4856,6 @@ def _create_model_from_files(
         ctfidf_config["vectorizer_model"]["params"]["ngram_range"] = tuple(ngram_range)
 
     params["n_gram_range"] = tuple(params["n_gram_range"])
-    ctfidf_config
 
     # Select HF model through SentenceTransformers
     try:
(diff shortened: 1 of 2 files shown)


hedgeho:
Not entirely sure why the error in checks appears, will come back to this later unless someone fixes it before that.

hedgeho:
Should work now. Tested saving&loading locally with environment without torch. 

Tested also loading of a model that was saved with stable version, the model behaved exactly the same.

MaartenGr:
Thanks for the PR, it is greatly appreciated. Just did some tests and it is working nicely for me. The code is looking great, so no comments there.

PR #2317, opened 2025-03-25
Update _bertopic.py to address #2316

### LINKED ISSUE #2316: Can't get updated topic_embeddings_ when using pre-calculated document embeddings?

### Desribe the bug

I can use the `update_topics` function by passing in the appropriate documents and my manually generated topic labels, and using the auto-generated embeddings. This works as intended and I get the updated `topic_embeddings_`.
`topic_model.update_topics(docs, topics=topics)`

However, if I calculate the document embeddings myself and pass them to BERTopic like so:
`topic_model.fit_transform(docs, embeddings=embeddings)`
then the `topic_embeddings_` are not updated after using `update_topics`. This is because the call to `_create_topic_vectors` within `update_topics` is done without any arguments being passed, and would normally use `self.embedding_model`, which is None when I define my own embeddings. 

How should I perform the update to get new `topic_embeddings_` in this case? 
Could a kwarg for passing pre-calculated embeddings to `update_topics` be added?

### BERTopic Version

Version: 0.16.4

nhansendev:
Added "embeddings" kwarg to update_topics for use when pre-calculated document embeddings are used. Addresses issue #2316

# What does this PR do?

Fixes #2316

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -1488,6 +1488,7 @@ class BERTopic:
         vectorizer_model: CountVectorizer = None,
         ctfidf_model: ClassTfidfTransformer = None,
         representation_model: BaseRepresentation = None,
+        embeddings: np.ndarray = None,
     ):
         """Updates the topic representation by recalculating c-TF-IDF with the new
         parameters as defined in this function.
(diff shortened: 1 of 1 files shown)


nhansendev:
The user could easily miss that they need to provide the embeddings to `update_topics`. Should there be some sort of warning?

MaartenGr:
Might indeed be nice to have such a check here but I'm not sure whether it would cover all cases considering the `_create_topic_vectors` also has a few different options for creating topic vectors.

MaartenGr:
I'm rerunning all jobs since the error didn't relate to your code but a connection issue with huggingface.

PR #2336, opened 2025-04-14
fix incorrect comparison in update_topics MaartenGr/BERTopic#2332

### LINKED ISSUE #2332: Potential bug: Incorrect type comparison in update_topics method
Hi,

First of all, thank you for this amazing library. While using BERTopic, I noticed what appears to be a logical error in the code.

In the `update_topics` method, there is a condition:

Image

However,  `self.topics_` is always a list, while the condition compares a **set** with a **list**. This comparison will always evaluate to True since a set and a list are different types, making the condition redundant.

Would you please confirm if this is indeed a bug? Let me know if you need any additional information.

uply23333:
# What does this PR do?
Fixes #2332

## Changes Made
Modified the topic comparison logic in `_bertopic.py` from using set comparison `set(topics) != self.topics_` to direct comparison `topics != self.topics_`. This change affects how topic updates are handled during the `update_topics()` method.

### Previous Behavior
- The set comparison `set(topics) != self.topics_` would trigger  unnecessary recalculations in `_create_topic_vectors` when the topic set remained the same

### Current Behavior
- The direct comparison `set(topics) != set(self.topics_)` is more precise
- Topic embedding recalculation only occurs when there's an actual change in topic set

## Impact on Testing
The change affects test cases in `test_bertopic.py` where `update_topics()` is called twice, which will lead to different result of `merge_models` 
        

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -1574,7 +1574,7 @@ class BERTopic:
         self.topic_representations_ = self._extract_words_per_topic(words, documents)
 
         # Update topic vectors
-        if set(topics) != self.topics_:
+        if set(topics) != set(self.topics_):
             # Remove outlier topic embedding if all that has changed is the outlier class
             same_position = all(
                 [
--- a/tests/test_bertopic.py
+++ b/tests/test_bertopic.py
@@ -152,5 +152,4 @@ def test_full_model(model, documents, request):
     topic_model1 = BERTopic.load("model_dir")
     merged_model = BERTopic.merge_models([topic_model, topic_model1])
 
-    assert len(merged_model.get_topic_info()) > len(topic_model1.get_topic_info())
     assert len(merged_model.get_topic_info()) > len(topic_model.get_topic_info())


uply23333:
Seems like this would affect `merge_topic`, i will figure it out.

uply23333:
> Seems like this would affect `merge_topic`, i will figure it out.
Hi MaartenGr，
The issue has been resolved. The test code required some adjustments. Also, the GitHub check failed earlier due to a network error with sklearn, so a re-run of the check might be necessary.

MaartenGr:
It seems that you removed a test to make it pass. Could you elaborate on why this was needed? I generally do not want to remove tests if they help uncover issues.

uply23333:
Thanks for your reply.

First of all, the comparison between `merged_model` and `topic_model1` in the test is invalid.
According to both the method description and the code logic, there is no guarantee that all topics in `topic_model1` will be retained in the merged result. The `Bertopic.merge_models` function uses cosine similarity to determine whether a topic from the second model is sufficiently similar to one in the baseline model. If not, it is considered a “new topic” and added to the merged result. Otherwise, it may be reassigned or dropped, which means the final topic list in `merged_model` can differ significantly from `topic_model1`.
```python
# Excerpt from Bertopic.merge_models
for tensors, selected_topics in zip(all_tensors[1:], all_topics[1:]):
    sim_matrix = cosine_similarity(tensors, merged_tensors)
    sims = np.max(sim_matrix, axis=1)

    new_topics = sorted(
        [index - selected_topics["_outliers"] for index, sim in enumerate(sims) if sim < min_similarity]
    )
```

---
Secondly, the original test passed due to a side effect in the old `update_topics` implementation.
The old logic would call `_create_topic_vectors`() twice, causing the `topic_embeddings_` to be recomputed even when the actual topics didn’t change. This behavior masked the issue and made the test pass incorrectly.
```python
# Excerpt from test_bertopic.py 
vectorizer_model = topic_model.vectorizer_model
topic_model.update_topics(documents, n_gram_range=(2, 2))

updated_topic = topic_model.get_topic(1)[:10]

topic_model.update_topics(documents, vectorizer_model=vectorizer_model)
original_topic = topic_model.get_topic(1)[:10]



# Excerpt from _bertopic.py update_topics
if set(topics) != self.topics_:
    ...
    else:
        self._create_topic_vectors()  # Triggered twice due to the test calling update_topics() twice


# Excerpt from _bertopic.py _create_topic_vectors
elif self.embedding_model is not None and type(self.embedding_model) is not BaseEmbedder:
   # Topic embeddings based on keyword representations
        
```
---
In summary, the test was based on an incorrect assumption about merge_models, and its previous passing behavior was due to unintended double recalculation of topic embeddings.

MaartenGr:
Thank you for taking the time to write this down. It is truly appreciated as I now understand the intention behind removing it (which was a nice catch btw). 

Everything else looks great, so I'll go ahead and merge this.

PR #2338, opened 2025-04-14
Fix/static embedding error

### LINKED ISSUE #2331: ImportError: cannot import name 'StaticEmbedding' from 'sentence_transformers.models'

### Desribe the bug

Title pretty much says it all.

```
BERTopic/bertopic/backend/_sentencetransformers.py:4
    from sentence_transformers.models import StaticEmbedding
```

Looks like this is an issue with versions. SentenceTransformer didn't add Model2Vec till
https://github.com/MaartenGr/BERTopic/pull/2245
but there's no requirement that the new version of SentenceTransformer is installed.

I'll submit a PR to fix this, but I don't know what solution you'd prefer. I think the best option is to move that import from line 4 to line 58, since it appears that's the only place it's needed. I'll make a PR and test it.

### Reproduction

```python
from bertopic import BERTopic

```

### BERTopic Version

v0.17.0

stevetracvc:
# What does this PR do?

Fixes import error if sentence_transformers isn't v3.2.0 or higher

Fixes #2331

## Before submitting

### DIFF
--- a/bertopic/backend/_sentencetransformers.py
+++ b/bertopic/backend/_sentencetransformers.py
@@ -1,7 +1,6 @@
 import numpy as np
 from typing import List, Union
 from sentence_transformers import SentenceTransformer
-from sentence_transformers.models import StaticEmbedding
 
 from bertopic.backend import BaseEmbedder
 
@@ -55,6 +54,7 @@ class SentenceTransformerBackend(BaseEmbedder):
 
         self._hf_model = None
         if model2vec and isinstance(embedding_model, str):
+            from sentence_transformers.models import StaticEmbedding
             static_embedding = StaticEmbedding.from_model2vec(embedding_model)
             self.embedding_model = SentenceTransformer(modules=[static_embedding])
         elif isinstance(embedding_model, SentenceTransformer):
--- a/bertopic/backend/_sentencetransformers.py
+++ b/bertopic/backend/_sentencetransformers.py
@@ -54,7 +54,15 @@ class SentenceTransformerBackend(BaseEmbedder):
 
         self._hf_model = None
         if model2vec and isinstance(embedding_model, str):
-            from sentence_transformers.models import StaticEmbedding
+            try:
+                from sentence_transformers.models import StaticEmbedding
+            except ImportError:
+                raise ValueError(
+                    "Cannot import StaticEmbedding from "
+                    "sentence_transformers. Please upgrade "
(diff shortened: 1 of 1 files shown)


stevetracvc:
Not sure whether it should be a ValueError or ImportError. Technically it's an ImportError but only because an invalid value was passes (`model2vec=True`)

MaartenGr:
That's an interesting question. I would also think it's closer to the parameter `model2vec=True` and as such would indeed consider ValueError over ImportError. In all honesty, they are both "correct" to me considering they both showcase an method of resolving the issue (using a different parameter `ValueError` versus upgrading sentence-transformers `ImportError`).

Other than that, it all looks good to me. Anything else to consider?

dschwalm:
I am facing the same issue and maybe it would be clearer to increase the min. version of sentence-transformers?

"sentence-transformers>=0.4.1" -> "sentence-transformers>=3.2.0"

3.2.0 where the StaticEmbedding class was added to sentence-transformers.

I understand that it may bring in additional dependencies, but clearer to me than moving this import under a conditional branch.

MaartenGr:
@dschwalm Ah, I have a tendency to keep dependencies as they are. Sorry! Going with `0.4.1` is perhaps a bit of a stretch here. I'm actually quite alright here with simply upping the version. That would indeed solve all issues and considering `sentence-transformers` is already at `>5.0` that seems like a good way forward.

dschwalm:
Totally makes sense. In our case we needed to stick with 0.16.4 as our dependencies could not be resolved in a combination that allows sentence-transformers  to be >=3.

MaartenGr:
@stevetracvc What do you think? Should we merely up `sentence-transformers` here? Especially with the many major updates it has had over the years I feel like we should look forward.

stevetracvc:
Sorry, lost track of this one. I generally just run my own forks anyway.

I don't upgrade packages very often, since it is an insane hassle to get everything I need into one environment (pytorch, tensorflow, cuML, and many others). I'm still using sentence-transformers 2.6.1 and bertopic 0.17.0. So we can close this and not use it if you want to push the requirement higher, but is there really a reason why it needed the higher version? I generally pre-run my embeddings since I store them in a Milvus server.

MaartenGr:
Upping the version helps maintain the stability of this package as well as prevent the need to maintain all older versions. I think upping the version here is the way to go, so this can indeed be closed.

PR #2354, opened 2025-05-12
Fix type annotations for _extract_representative_docs

ldorigo:
# Small type annotation fix

There's a lot of incorrect type annotations in the library - I will submit a few small PRs for the ones that are directly relevant to me (as they caused confusion and type issues in my own code).


MaartenGr:
I believe many of those that you mentioned should be fixed in https://github.com/MaartenGr/BERTopic/pull/2457

If not, feel free to re-open this PR.

PR #2378, opened 2025-06-11
Update _bertopic.py

Guyniran:
Minor change to documentation.
HDBSCAN is the default clustering algorithm. So it was confusing whether min_topic_size will be used by it or not.


MaartenGr:
I'm not sure if this clarifies things. The `hdbscan_model` that is mentioned refers to the actual parameters:

```python
BERTopic(hdbscan_model=...)
```

In other words, if you use that **parameter** then `min_topic_size` will not be used regardless of cluster model.

PR #2401, opened 2025-07-18
Allow execution without plotly

### LINKED ISSUE #2398: Allow execution without plotly
### Feature request

Would it be possible to include an execution/installation option where plotly is not required? 

This could be in the form of a lightweight no-deps installation like suggested for umap and dbscan:

https://maartengr.github.io/BERTopic/getting_started/tips_and_tricks/tips_and_tricks.html#lightweight-installation

Currently, trying to do this, the code crashes because line 56 of bertopic/_bertopic.py  is importing plotting module that itselg import plotly.

### Motivation

When I am using bertopic trained model for inference I don't need plotting, the same way I don't need umap or hdbscan, see the lightweight installation option.

### Your contribution

If approved, I can raise a PR with a simple fix such as catching the exception in _bertopic.py with a clear error signal. Then it's up to the user responsibility not to call the visualize methods of BERTopic - the user has already "hacked" the environment to remove plotly which is a required dependency.

Thanks!

luismavs:
Allow execution without plotly.

Fixes #2398

Handles calls to .visualize... methods with a ```pass``` and a logger.warning

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -53,7 +53,6 @@ from sklearn.metrics.pairwise import cosine_similarity
 from sklearn.feature_extraction.text import CountVectorizer, TfidfTransformer
 
 # BERTopic
-from bertopic import plotting
 from bertopic.cluster import BaseCluster
 from bertopic.backend import BaseEmbedder
 from bertopic.representation._mmr import mmr
@@ -73,11 +72,22 @@ from bertopic._utils import (
 )
 import bertopic._save_utils as save_utils
 
-# Visualization
-import plotly.graph_objects as go
 
 logger = MyLogger()
 logger.configure("WARNING")
+
+try:
+    from bertopic import plotting
+    import plotly.graph_objects as go
+
+except ModuleNotFoundError as e:
+    if "No module named 'plotly'" in str(e):
+        logger.warning("Plotly is not installed. Please install it to use the plotting functions.")
+        from bertopic._utils import mock_plotly_go as go, MockPlotting
+
+        plotting = MockPlotting(logger)
+    else:
+        raise ModuleNotFoundError(e)
 
 
 class BERTopic:
--- a/bertopic/_utils.py
+++ b/bertopic/_utils.py
@@ -226,3 +226,28 @@ def select_topic_representation(
             repr_, ctfidf_used = embeddings, False
 
     return to_ndarray(repr_) if output_ndarray else repr_, ctfidf_used
+
+
+# Visualization mocks in case plotly is not installed
+class MockPlotting:
+    """Mock plotting module when plotly is not installed."""
+
+    def __init__(self, logger: MyLogger):
+        self.logger = logger
+
+    def __getattr__(self, name):
+        def mock_function(*args, **kwargs):
+            self.logger.warning(f"Plotly is not installed. Cannot use {name} visualization function.")
+            return MockFigure()
+
+        return mock_function
+
+
+class MockFigure:
+    """Mock class for plotly.graph_objects.Figure when plotly is not installed."""
+
+    def __init__(self, *args, **kwargs):
+        pass
+
+
+mock_plotly_go = type("MockPlotly", (), {"Figure": MockFigure})()
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,13 @@
 test:
 	pytest
 
+test-no-plotly:
+	uv sync --extra test
+	uv pip uninstall plotly
+	pytest tests/test_other.py -k plotly
+	uv sync --extra test
+	pytest tests/test_other.py -k plotly
+
 coverage:
 	pytest --cov
 
--- a/tests/test_other.py
+++ b/tests/test_other.py
@@ -1,4 +1,14 @@
 from bertopic import BERTopic
+from bertopic.dimensionality import BaseDimensionalityReduction
+
+try:
+    import plotly.graph_objects as go
+
+    figure_type = go.Figure
+except ImportError:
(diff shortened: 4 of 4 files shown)


bertopic/_bertopic.py
```diff
@@ -73,12 +72,23 @@
...
 logger = MyLogger()
 logger.configure("WARNING")
 
+try:
+    from bertopic import plotting
+    import plotly.graph_objects as go
+
+except ModuleNotFoundError as e:
+    if "No module named 'plotly'" in str(e):
+        logger.warning("Plotly is not installed. Please install it to use the plotting functions.")
```
MaartenGr, line 85: Perhaps also add a way to install it? A one-liner would suffice, something like "For example with `pip install plotly` or `uv pip install plotly`". That way, users would not have to search how to install plotly but can do it immediately.

bertopic/_bertopic.py
```diff
@@ -73,12 +72,23 @@
...
 
 logger = MyLogger()
 logger.configure("WARNING")
 
+try:
+    from bertopic import plotting
+    import plotly.graph_objects as go
+
+except ModuleNotFoundError as e:
+    if "No module named 'plotly'" in str(e):
```
MaartenGr, line 84: I'm wondering whether this is a good enough check to see whether plotly is installed. Checking the string might be unstable. Perhaps use `importlib` instead?

```python
plotly_available = importlib.util.find_spec("plotly") is not None
```

bertopic/_utils.py
```diff
@@ -226,3 +226,28 @@ def to_ndarray(array: Union[np.ndarray, csr_matrix]) -> np.ndarray:
...
+
+    def __init__(self, logger: MyLogger):
+        self.logger = logger
+
+    def __getattr__(self, name):
+        def mock_function(*args, **kwargs):
+            self.logger.warning(f"Plotly is not installed. Cannot use {name} visualization function.")
+            return MockFigure()
+
+        return mock_function
```
MaartenGr, line 243: Let's simplify this by raising an error instead.

bertopic/_utils.py
```diff
@@ -226,3 +226,28 @@ def to_ndarray(array: Union[np.ndarray, csr_matrix]) -> np.ndarray:
...
+
+
+class MockFigure:
+    """Mock class for plotly.graph_objects.Figure when plotly is not installed."""
+
+    def __init__(self, *args, **kwargs):
+        pass
```
MaartenGr, line 250: I don't think this is necessary when we can use `def some_function(self) -> "go.Figure"` instead. This would reduce the code quite a bit.

MaartenGr (commented):
Thank you for the PR! I left some comments here and there. Most importantly, I think this solution is more complex than is currently needed. We only need to mock the plotting as we can do the type-hinting of `go.Figure` like so: `"go.Figure"`. 

As an example, we can reduce the complexity to this:

```python
import importlib.util
import logging
from typing import Any


class MockPlotlyModule:
    """Mock module that raises informative errors when plotly functions are called."""
  
    def __getattr__(self, name: str) -> Any:
        def mock_function(*args, **kwargs):
            raise ImportError(
                f"Plotly is required to use '{name}'. "
                "Install it with: pip install plotly"
            )
        return mock_function

# Clean import logic using importlib
if importlib.util.find_spec("plotly") is None:
    logger.warning("Plotly is not installed. Plotting functions will raise errors when called.")
    
    # Mock modules
    plotting, go = MockPlotlyModule(), None
    
else:
    from bertopic import plotting
    import plotly.graph_objects as go
```

I'm also wondering if we should do the logger.warning here as we cannot control that verbosity. There might be use-cases where you do not want that logging.

luismavs:
> Thank you for the PR! I left some comments here and there. Most importantly, I think this solution is more complex than is currently needed. We only need to mock the plotting as we can do the type-hinting of `go.Figure` like so: `"go.Figure"`.
> 
> As an example, we can reduce the complexity to this:
> 
> ```python
> import importlib.util
> import logging
> from typing import Any
> 
> 
> class MockPlotlyModule:
>     """Mock module that raises informative errors when plotly functions are called."""
>   
>     def __getattr__(self, name: str) -> Any:
>         def mock_function(*args, **kwargs):
>             raise ImportError(
>                 f"Plotly is required to use '{name}'. "
>                 "Install it with: pip install plotly"
>             )
>         return mock_function
> 

I like this overall approach, simpler indeed.

The visualize methods now return string literal type hint "go.Figure" which works fine (as far as the type checker goes) in both cases - with and without plotly installed.

I've also gone ahead and fixed ```visualize_document_datamap``` method too, to have return type hint matplotlib Figure

> # Clean import logic using importlib
> if importlib.util.find_spec("plotly") is None:
>     logger.warning("Plotly is not installed. Plotting functions will raise errors when called.")

Fine. For consistency, the check for HDBSCAN on line 41 should use something similar.

>     # Mock modules
>     plotting, go = MockPlotlyModule(), None
>     
> else:
>     from bertopic import plotting
>     import plotly.graph_objects as go
> ```
> 
> I'm also wondering if we should do the logger.warning here as we cannot control that verbosity. There might be use-cases where you do not want that logging.

I'm ok with removing that logger.warning - there is already a clear exception if the user tries to use visualizations without plotly present. Also, ideally BERTopic should have a way to configure the log level more dynamically rather than the hardcoded WARNING level.

MaartenGr:
Apologies for the delay and thank you for the updated code. I agree with the logger, it should be handled a bit more elegantly than it is now (mostly handled in the main class and not easily changed). Either way, I'll go ahead and merge this, so thank you for your work on this. It is greatly appreciated!

PR #2412, opened 2025-08-25
Make UMAP's init default to random on visualize_topics for reproducible visualization

### LINKED ISSUE #2403: Visualize_topics randomness

### Desribe the bug

Thank you for your great package sir,

I just found a bug, it seems that even if I have set the random state in UMAP's parameter, the visualization of LDAvis's bubbles is totally random. The function .get_topic_info() is reproducible, but no matter what I try for the .visualize_topics(), it just does not work. I checked your _topics.py file, it looks like you also set the random state for the visualization in terms of UMAP, so I wonder whether the bubble chart could be reproducible, or I just need to save it in case the results vanished. 

### Reproduction

```python
from bertopic import BERTopic

```

### BERTopic Version

0.15.0

makramab:
# What does this PR do?

Fixes #2403

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -2548,7 +2548,6 @@ class BERTopic:
         topics: List[int] = None,
         top_n_topics: int = None,
         use_ctfidf: bool = False,
-        umap_init: str = "random",
         custom_labels: bool = False,
         title: str = "<b>Intertopic Distance Map</b>",
         width: int = 650,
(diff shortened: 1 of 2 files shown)


MaartenGr:
Thank you for the PR! 

At the moment, I'm **very** hesistant to add another parameter to the BERTopic class. I'd like to keep it all as small as possible with modularity as an important driver. As such, I would prefer to prevent model-specific parameters. Instead, I would suggest setting the init in `visualize_topics` to "random" as that should be the only (one of the few?) place(s) where it is used.

makramab:
Thanks for the feedback! That makes sense. I’ll go ahead and change the visualize_topics function to set init="random" instead of adding a parameter.

makramab:
Just pushed the changes, @MaartenGr.  thanks again for the review!

MaartenGr:
Wow, look at me waiting 3 weeks before answering... Sorry! Thank you for the changes and your patience. Looks good to me and can be merged.

Issue #2335, opened 2025-04-14
Adding support for cuML's nn-descent with UMAP


jemather:
### Feature request

Adding support for nearest neighbor descent when using cuML UMAP

### Motivation

RAPIDS cuML added support for nearest-neighbor descent in UMAP, which approximates the full nearest-neighbor ("brute force") graph ([NVidia has a good explainer here](https://developer.nvidia.com/blog/even-faster-and-more-scalable-umap-on-the-gpu-with-rapids-cuml/)). This makes it ***much*** faster to run UMAP, and in my use case made it feasible to even run UMAP, as I ran out of memory with the brute force approach.

However, this isn't an option that can be set when initializing the model and passing it to BERTopic. Instead, it requires running fit_transform as a single step and including the "data_on_host=True" parameter. But the current version of BERTopic separates fit and transform in _reduce_dimensionality. I just monkey patched it myself and got it working, but I think fixing it would be relatively easy for the full model fit. I'm unsure if it would work with a partial fit, though, so I did not implement it with partial_fit. 

### Your contribution

```
    def _reduce_dimensionality(
        self,
        embeddings: Union[np.ndarray, csr_matrix],
        y: Union[List[int], np.ndarray] = None,
        partial_fit: bool = False,
        data_on_host = True # only compatible with using nn_descent, which is an option set in the UMAP model, so setting this may be more complex than what I am doing here.
    ) -> np.ndarray:
        """Reduce dimensionality of embeddings using UMAP and train a UMAP model.

        Arguments:
            embeddings: The extracted embeddings using the sentence transformer module.
            y: The target class for (semi)-supervised dimensionality reduction
            partial_fit: Whether to run `partial_fit` for online learning

        Returns:
            umap_embeddings: The reduced embeddings
        """
        logger.info("Dimensionality - Fitting the dimensionality reduction algorithm")
        # Partial fit
        if partial_fit:
            if hasattr(self.umap_model, "partial_fit"):
                self.umap_model = self.umap_model.partial_fit(embeddings)
            elif self.topic_representations_ is None:
                self.umap_model.fit(embeddings)
            umap_embeddings = self.umap_model.transform(embeddings)

        # Regular fit
        else:
            try:
                # cuml umap needs y to be an numpy array
                y = np.array(y) if y is not None else None
                umap_embeddings = self.umap_model.fit_transform(embeddings, y=y, data_on_host)
            except TypeError:
[...]

PR #2416, opened 2025-08-29
 Modify `_reduce_dimensionality` to use `fit_transform`

betatim:
# What does this PR do?

This picks up #2347, which modifies `_reduce_dimensionality` to use `fit_transform` when possible. This makes it possible to use nearest neighbor descent with cuML UMAP.

Fixes https://github.com/MaartenGr/BERTopic/issues/2335

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -3784,7 +3784,7 @@ class BERTopic:
             if hasattr(self.umap_model, "partial_fit"):
                 self.umap_model = self.umap_model.partial_fit(embeddings)
                 umap_embeddings = self.umap_model.transform(embeddings)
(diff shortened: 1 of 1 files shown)


betatim:
Thanks for the code suggestion, I applied it. Also added `y` as argument.

This time it was my turn to be slow to reply :D

MaartenGr:
@betatim Thanks for the help on this! I'll do something about those duplicate `umap_embeddings = topic_model.transform(embeddings)` in a different PR but for now this can be merged.

betatim:
Thanks a lot for the help and patience!

I agree the duplicates are not nice, but I don't have a great idea right now. Feels like it needs abstracting away somehow :-/

PR #2417, opened 2025-09-06
Fix topic_sizes_ not updated in zero-shot topic modeling (#2384)

### LINKED ISSUE #2384: Zero-shot topic modeling with nr_topics parameter results in empty topic_sizes_ and get_topic_info()

### Desribe the bug

When using zero-shot topic modeling with the `nr_topics` parameter set, and all (or most) documents are assigned to zero-shot topics, the `topic_sizes_` attribute remains empty and `get_topic_info()` returns an empty DataFrame.

**Root cause:**
The issue occurs in the `fit_transform()` method when all documents are assigned to zero-shot topics. The code has this conditional:

```python
if len(documents) > 0:
    # Cluster reduced embeddings
    documents, probabilities = self._cluster_embeddings(umap_embeddings, documents, y=y)
        documents, embeddings = self._combine_zeroshot_topics(
            documents, embeddings, assigned_documents, assigned_embeddings
        )
else:
```

When all documents are assigned to zero-shot topics, `len(documents) == 0`, so the `else` branch is taken. However, this branch never calls `_update_topic_size()`, which is normally called within `_cluster_embeddings()` or `_combine_zeroshot_topics()`.

Additionally, when `nr_topics` is specified, the fallback call to `_sort_mappings_by_frequency()` (which would call `_update_topic_size()`) is skipped because `if not self.nr_topics:` evaluates to `False`.

**Expected behavior:**
- `topic_sizes_` should contain the count of documents per topic
- `get_topic_info()` should return a populated DataFrame with topic information

**Actual behavior:**
- `topic_sizes_` is an empty dictionary `{}`
- `get_topic_info()` returns an empty DataFrame
- `topic_representations_` works correctly (populated as expected)

**Note:** The issue does NOT occur when `nr_topics` is not specified, because `_sort_mappings_by_frequency()` gets called, which internally calls `_update_topic_size()`.

### Reproduction

```python
from bertopic import BERTopic

# Sample documents and zero-shot topics
docs = ["I need help with my voucher", "Gift card not working", "Customer service was poor"] * 50
zeroshot_topics = ["Voucher inquiries", "Gift card issues", "Customer service feedback"]

# BUG: Setting nr_topics causes the issue
model = BERTopic(
    zeroshot_topic_list=zeroshot_topics,
    zeroshot_min_similarity=0.0,  # Force all documents to zero-shot assignment
    nr_topics=4,  # This triggers the bug
)

topics, _ = model.fit_transform(docs)

# Demonstrates the bug
print("topic_sizes_:", model.topic_sizes_)  # Expected: Counter with topic counts, Actual: None
print("topic_representations_ works:", bool(model.topic_representations_))  # This works: True

```

### BERTopic Version

0.17.0

ck37:
# What does this PR do?

Minor fix to fit_transform() to fix the issue of  `topic_sizes_` not being updated in zero-shot topic modeling when using the `nr_topics` parameter. Adds tests to confirm the fix and provides a small changelog update.

Happy to revise further if any revisions are desired.

Fixes #2384

## Before submitting


MaartenGr:
Thank you for the PR! With respect to the tests, can you explain a bit more on why those were added like this? Specifically, I'm seeing UMAP parameters sometimes being added whilst that is not the case for every test. Moreover, there are several calls to BERTopic, which can be quite expensive in a testing pipeline especially if you could run it once instead. There are conftest available where you can initialize a given model once and re-use it across tests:

https://github.com/MaartenGr/BERTopic/blob/master/tests/conftest.py

PR #2443, opened 2025-10-16
chore: update pre-commit hooks

afuetterer:
# What does this PR do?

This PR bumps the pre-commit hooks again and uses the new ruff-check id, silencing a deprecation warning when running the hook.

## Before submitting

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ spacy = [
 test = [
     "pytest>=5.4.3",
     "pytest-cov>=2.6.1",
-    "ruff~=0.4.7",
+    "ruff~=0.14.0",
 ]
 use = [
     "tensorflow",


afuetterer:
@MaartenGr: A question about your workflow. 

I saw the commands in the Makefile

```
format:
	ruff format
lint:
	ruff check --fix
```

Do you use these commands regularly? The version is used here is defined in pyproject.toml like this `ruff~=0.4.7`. This leads to 0.4.10 in a fresh install today. It is different from the version in pre-commit-config.yaml.

I never run ruff manually like this, but rely on the pre-commit hook.

But you could use `pre-commit run ruff-check` and `pre-commit run ruff-format` in the Makefile? Then you would have only one point to define the ruff version.

What do you think?

MaartenGr:
> Do you use these commands regularly? The version is used here is defined in pyproject.toml like this ruff~=0.4.7. This leads to 0.4.10 in a fresh install today. It is different from the version in pre-commit-config.yaml.

Not at all. I used to and have since switched to pre-commit (although that has been giving me problems in the past when installing in different environments). 

That said, I want to switch over to [`Justfile`](https://github.com/casey/just) over `Makefile` as those features are easier to use across platforms. It also might make it easier to run it 

> I never run ruff manually like this, but rely on the pre-commit hook.

I also have it run automatically in VSCode, which I believe is based on the plugin. 

> But you could use pre-commit run ruff-check and pre-commit run ruff-format in the Makefile? Then you would have only one point to define the ruff version.

That would move the dependency management of `ruff` in `pyproject.toml` to `.pre-commit-config.yaml`, right? That's an interesting step... Intuitively, I would say that it should belong in `pyproject.toml`, especially if people are getting their `ruff` versioning from the .venv and used automatically in VSCode, for instance.

afuetterer:
> That would move the dependency management of `ruff` in `pyproject.toml` to `.pre-commit-config.yaml`, right? 

Yes, that would be the case.

> That's an interesting step... Intuitively, I would say that it should belong in `pyproject.toml`, especially if people are getting their `ruff` versioning from the .venv and used automatically in VSCode, for instance.

It is not necessary of course. If `ruff` would and should (?) be run manually via the `ruff` command (or in the IDE), it needs to be defined in pyproject.toml. I just wanted to point the difference 0.4.0 (test dependency version) vs 0.14.0 (pre-commit hook version). In "ruff years" that is a huge difference.

Your project, you decide. Bump in pyproject.toml as well?

afuetterer:
> That said, I want to switch over to [`Justfile`](https://github.com/casey/just) over `Makefile` as those features are easier to use across platforms. It also might make it easier to run it

Yeah, I prefer justfiles over Makefiles as well.

MaartenGr:
Yes, let's bump in pyproject.toml.

afuetterer:
Done.

PR #2445, opened 2025-10-23
ci: switch to uv based project installation

afuetterer:
# What does this PR do?

This is a new version of #2082. The PR adds the `setup-uv` GHA to install uv and setup the python version from the matrix.

To still use the `make check`, I needed to add `activate-environment: true`, otherwise there is no pytest in PATH. Another solution would to call `uv run pytest` instead of `make check`.

## Before submitting

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -36,5 +36,7 @@ jobs:
         activate-environment: "true"
     - name: Install dependencies (including test group)
       run: uv sync --extra test
+    - name: Display the project's dependency tree (including information on outdated packages)
+      run: uv tree --outdated
     - name: Run Checking Mechanisms
       run: make check


afuetterer:
Install step is already a lot faster.

afuetterer:
~Need to add `enable-cache: "false"`.~

Edit: I checked in my fork. setup-uv caches the uv installation per default settings, it runs prune cache per default settings as well. So the cache does not include downloaded pypi packages (which would be huge for this project).

```
/opt/hostedtoolcache/uv/0.9.5/x86_64/uv cache prune --ci --force
Pruning cache at: /home/runner/work/_temp/setup-uv-cache
Removed 28282 files (7.1GiB)
```
Ref: https://github.com/MaartenGr/BERTopic/actions/runs/18743018909/job/53463656594

So the caching can stay enabled as "auto", I guess.

This switch to setup-uv gets you ~2 minutes faster CI times. I think it's a win.

afuetterer:
I also added a `uv tree --outdated` to the CI run. It adds 0 seconds to the time.

This might be useful for debugging which packages / dependencies are installed in CI to pinpoint issues due to dependencies.

What do you think?

Truncated output:

```
Run uv tree --outdated
Resolved 228 packages in 0.78ms
bertopic v0.17.3
├── hdbscan v0.8.40
│   ├── joblib v1.5.2
│   ├── numpy v2.1.3 (latest: v2.2.6)
│   ├── scikit-learn v1.7.2
│   │   ├── joblib v1.5.2
│   │   ├── numpy v2.1.3
│   │   ├── scipy v1.15.3
│   │   │   └── numpy v2.1.3
│   │   └── threadpoolctl v3.6.0
│   └── scipy v1.15.3 (*)
...
```

MaartenGr:
Good idea adding the tree here! That is going to help so much with debugging all of this. Nice features all around, thank you for your work on this.

PR #2446, opened 2025-10-24
Update installation instructions for cuML with BERTopic.

csadorf:
# What does this PR do?

Updates cuML installation instructions across the documentation to use simplified PyPI.org installation (available as of cuML 25.10) with current CUDA versions.

### Updated Installation Instructions
Updated cuML + BERTopic installation instructions in three documentation files:
- `docs/getting_started/clustering/clustering.md`
- `docs/getting_started/dim_reduction/dim_reduction.md`
- `docs/getting_started/tips_and_tricks/tips_and_tricks.md`

### Key Updates
1. **Simplified PyPI.org installation**: Uses direct pip install from PyPI.org instead of requiring nvidia.com package indexes with `--extra-index-url` flags
2. **Updated CUDA versions**: Replaced outdated CUDA 11 packages with current CUDA 12 and CUDA 13 options
3. **Streamlined dependencies**: Removed auxiliary packages (cudf, dask-cudf, cugraph, cupy) that are not required for basic cuML usage with BERTopic or automatically installed as dependencies
4. **Recommended installation order**: Install cuML first, then BERTopic to avoid pip resolver dependency conflicts with CUDA runtime packages
5. **Added helpful context**:
   - Warning about pip resolver limitations
   - Note that cuML is pre-installed on Google Colab
   - Link to RAPIDS installation guide for detailed/advanced installation instructions

### Motivation
The updated instructions use pip install from PyPI.org, which is now possible as of cuML release version 25.10. This simplifies the installation process by removing the need for additional package indexes (`--extra-index-url`) and auxiliary packages (cudf, dask-cudf, cugraph, cupy).

Additionally, the two-step installation approach (install cuML first, then BERTopic) avoids pip resolver limitations when handling conflicting CUDA runtime dependencies between cuML and PyTorch (via sentence-transformers).

## Before submitting
- [x] This PR fixes a typo or improves the docs (if yes, ignore all other checks!).

### DIFF
--- a/docs/getting_started/clustering/clustering.md
+++ b/docs/getting_started/clustering/clustering.md
@@ -110,12 +110,23 @@ The great thing about using cuML's HDBSCAN implementation is that it supports ma
     However, it is still possible to calculate the topic-document probability matrix for the data on which the model was trained (i.e., `.fit` and `.fit_transform`).
 
 !!! note
-    If you want to install cuML together with BERTopic using Google Colab, you can run the following code:
+    To install cuML with BERTopic, run these commands:
 
+    **For CUDA 12:**
     ```bash
+    !pip install cuml-cu12
     !pip install bertopic
-    !pip install cudf-cu11 dask-cudf-cu11 --extra-index-url=https://pypi.nvidia.com
-    !pip install cuml-cu11 --extra-index-url=https://pypi.nvidia.com
-    !pip install cugraph-cu11 --extra-index-url=https://pypi.nvidia.com
-    !pip install --upgrade cupy-cuda11x -f https://pip.cupy.dev/aarch64
     ```
+
+    **For CUDA 13:**
+    ```bash
+    !pip install cuml-cu13
+    !pip install bertopic
+    ```
+
+    !!! warning
+        Install cuML first, then BERTopic. Installing both in a single command can fail due to pip resolver limitations with CUDA runtime dependencies.
+
+    **Note:** cuML is already installed on Google Colab.
+
+    For more detailed information on installing cuML, including additional dependencies and platform-specific instructions, see the [RAPIDS installation guide](https://docs.rapids.ai/install/).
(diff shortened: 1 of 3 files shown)


csadorf:
See https://github.com/rapidsai/cuml/issues/7374 for details on why we can't install cuml and bertopic with pip in one command. If we get that resolved, I can update the instructions here or we can leave them as-is.

MaartenGr:
Thank you for the PR and extensive descriptions, that helps me understand the context here a bit more. 

I'm also wondering that considering many are moving away from `pip` to `uv`, would it make sense to add instructions for `uv`and potentially fix the dependency conflicts by either pinning indices in `pyproject.toml` (which seems a lot of maintenance) or potentially add the simple [instruction](https://docs.astral.sh/uv/guides/integration/pytorch/#configuring-accelerators-with-environment-markers):

```
uv pip install torch --torch-backend=auto
```

What do you think?

csadorf:
> Thank you for the PR and extensive descriptions, that helps me understand the context here a bit more.
> 
> I'm also wondering that considering many are moving away from `pip` to `uv`, would it make sense to add instructions for `uv`and potentially fix the dependency conflicts by either pinning indices in `pyproject.toml` (which seems a lot of maintenance) or potentially add the simple [instruction](https://docs.astral.sh/uv/guides/integration/pytorch/#configuring-accelerators-with-environment-markers):
> 
> ```
> uv pip install torch --torch-backend=auto
> ```
> 
> What do you think?

We don't currently support uv installation of cuml. That is not to say it won't work. I'd rather stick to the currently documented approach and then update it if we find that uv installation works reliably or even better, RAPIDS officially supports it.

csadorf:
@MaartenGr Any chance we could merge this soon or do you think we need to make more changes? Also happy to address those in a follow-up.

MaartenGr:
@csadorf Apologies for the delay! Yes, sounds good to me.

csadorf:
> @csadorf Apologies for the delay! Yes, sounds good to me.

Thanks a lot!

PR #2451, opened 2025-10-29
Update topic mapping when all documents match zero-shot topics

### LINKED ISSUE #2447: Topic names not set to zero-shot topic when `fit_transform` results in only zero-shot topics

### Desribe the bug

The names for the topics are not set to the zero-shot topic names when the `fit_transform` results in only zero-shot topics .

If the result contains zero-shot topics and non-zero-shot topics the zero-shot topics get their name from `zeroshot_topic_list`. However when there are only zero-shot topics (by using a very low `zeroshot_min_similarity` for example) all documents get assigned to zero-shot topics, but the names are not taken from the list.

**Name for topic when running reproduction code:** 0_optimization_prediction_algorithms_optimal 
**Expected name for topic:** Clustering

[Looking at the code](https://github.com/MaartenGr/BERTopic/blob/0bed25cdac0445fe01869e47b3b0beeaf124230e/bertopic/_bertopic.py#L489C1-L499C45), when not all documents are assigned to zeroshot topics `len(documents) > 0` is true and `combine_zeroshot_topics` is called where the topic/label mapping is updated. When all documents have been assigned this function is not called and the mapping is not updated. 
```python
        if len(documents) > 0:
            # Cluster reduced embeddings
            documents, probabilities = self._cluster_embeddings(umap_embeddings, documents, y=y)
                documents, embeddings = self._combine_zeroshot_topics(
                    documents, embeddings, assigned_documents, assigned_embeddings
                )
        else:
```

### Reproduction

```python

from bertopic import BERTopic

# We select a subsample of 100 abstracts from ArXiv
docs = dataset["abstract"][:100]

zeroshot_topic_list = ["Clustering"]

# We fit our model using the zero-shot topics
# for clustering instead.
topic_model = BERTopic(
    min_topic_size=15,
    zeroshot_min_similarity=.000001, # Low value to make sure all documents are assigned to zero-shot topics
    representation_model=KeyBERTInspired()
)
topics, _ = topic_model.fit_transform(docs)
print(topic_model.get_topic_info())

```

### BERTopic Version

0.17.3

huisman:
Fix topic labels when all documents match zero-shot topics by updating the topic id mapping.

# What does this PR do?

When all documents match zero-shot topics, the labels for the topics are not set to the zero-shot topic name. 
This PR tries to fix this by updating the topic id/mapping and updating the topic sizes when there are no non-zero-shot topics.

Fixes #2447 

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -504,7 +504,7 @@ class BERTopic:
             documents = assigned_documents
             embeddings = assigned_embeddings
 
-             # Update topic sizes
+            # Update topic sizes
             self._update_topic_size(documents)
 
             # All documents matches zero-shot topics


MaartenGr:
Thank you for the PR! Could you perhaps give an example of how it fixes the issue? That way, I can run some tests. 

Also, did you test whether the predictions are still correct? I wonder whether the update topics would properly propagate the predictions through the mapping.

Lastly, I think you duplicated this:

```python
```

Which might break things depending on how it was ordered.

Issue #2450, opened 2025-10-28
Enable additional ruff rules


afuetterer:
I would like to propose enabling additional ruff rules.

There might be performance gains when enabling some of these? E.g.: 

- [x] RUF: https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf
- [x] PD: https://docs.astral.sh/ruff/rules/#pandas-vet-pd
- [ ] UP: https://docs.astral.sh/ruff/rules/#pyupgrade-up
- [ ] NPY: https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy

What do you think?

MaartenGr:
These seem interesting. I'm wondering how many issues it will flag and as a result how many should be fixed.

Are there any in particular that you think would be helpful?

afuetterer:
Let me get back to you with some numbers. The RUF rule has 100+ violations, but they are autofixable. So large number, but small effort.

afuetterer:
Adding these `"RUF",  "UP",  "NPY",  "PD"` and running `check` I get the following numbers:
```
$ ruff check --statistics
200     UP006   [*] non-pep585-annotation
116     RUF013  [ ] implicit-optional
 70     UP035   [ ] deprecated-import
 62     UP007   [*] non-pep604-annotation-union
 25     PD011   [ ] pandas-use-of-dot-values
 18     RUF015  [ ] unnecessary-iterable-allocation-for-first-element
 10     NPY002  [ ] numpy-legacy-random
  3     RUF022  [*] unsorted-dunder-all
  3     UP008   [ ] super-call-with-parameters
  2     RUF005  [ ] collection-literal-concatenation
  2     UP031   [ ] printf-string-formatting
  2     UP045   [*] non-pep604-annotation-optional
  1     PD015   [ ] pandas-use-of-pd-merge
  1     RUF010  [*] explicit-f-string-type-conversion
  1     RUF059  [ ] unused-unpacked-variable
  1     UP004   [*] useless-object-inheritance
  1     UP015   [*] redundant-open-modes
  1     UP032   [*] f-string
  1     UP034   [*] extraneous-parentheses
Found 520 errors.
[*] 288 fixable with the `--fix` option (142 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

I would like to tackle a few of those.

PR #2457, opened 2025-11-12
Enable ruff rule `RUF`

afuetterer:
# What does this PR do?

Discussion at #2450

1. Mostly it is something like this:

```python
nr_topics: Union[int, str] = None,
```
-->

```python
nr_topics: Union[int, str] | None = None,
```

If you add the type hint `Union[int, str]` you should not use a default value of None.

2. It also sorts entries in `__all__` alphabetically.

3. This `RUF015 unnecessary-iterable-allocation-for-first-element` might actually be a little faster. You don't need to convert result of zip to list to access first entry.

Summary:
```
$ ruff check --statistics
116     RUF013  [ ] implicit-optional
 18     RUF015  [ ] unnecessary-iterable-allocation-for-first-element
  3     RUF022  [*] unsorted-dunder-all
  2     RUF005  [ ] collection-literal-concatenation
  1     RUF010  [*] explicit-f-string-type-conversion
  1     RUF059  [ ] unused-unpacked-variable
```

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -148,11 +148,11 @@ class BERTopic:
         top_n_words: int = 10,
         n_gram_range: Tuple[int, int] = (1, 1),
         min_topic_size: int = 10,
-        nr_topics: Union[int, str] = None,
+        nr_topics: Union[int, str] | None = None,
         low_memory: bool = False,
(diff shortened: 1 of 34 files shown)


MaartenGr:
Looks great to me! Thanks again for the PR.

PR #2480, opened 2026-04-07
Fix TopicMapper.add_new_topics inserting None placeholders that break model.save

### LINKED ISSUE #2432: `model.save()` fails with `TypeError` when `topic_mapper_.mappings_` contains `None`

### Describe the bug

When calling `TopicMapper.add_new_topics`, the method appends new rows to `mappings_` with `None` placeholders:

```
def add_new_topics(self, mappings: Mapping[int, int]):
    length = len(self.mappings_[0])
    for key, value in mappings.items():
        to_append = [key] + ([None] * (length - 2)) + [value]
        self.mappings_.append(to_append)
```

This works during runtime, but later when saving the model with `model.save()` (serialization="safetensors"), it fails because `_save_utils.save_topics` tries to cast the entire `mappings_` table to `np.array(..., dtype=int)`:

```
File ".../bertopic/_save_utils.py", line 442, in save_topics
    "topic_mapper": np.array(model.topic_mapper_.mappings_, dtype=int).tolist(),
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

### Reproduction

```python
from sklearn.cluster import Birch
from bertopic.vectorizers import OnlineCountVectorizer
from bertopic import BERTopic

# Prepare documents
all_docs = fetch_20newsgroups(subset="train",  remove=('headers', 'footers', 'quotes'))["data"]
first_docs = all_docs[:50] # Making it small so that new clusters emerge with partial_fit and new mappings are added to the topic mapper
doc_chunks = [all_docs[50:][i:i+1000] for i in range(0, len(all_docs[50:]), 1000)]

# Prepare sub-models that support online learning
cluster_model = Birch(threshold=1.5, n_clusters=None)
vectorizer_model = OnlineCountVectorizer(stop_words="english", decay=.01),

# Train model for a first time
topic_model = BERTopic(
    language="multilingual",
    hdbscan_model=Birch(threshold=1.5, n_clusters=None),
    vectorizer_model = OnlineCountVectorizer(stop_words="english", decay=.01),
)

topic_model.fit_transform(documents=first_docs)

# Incremental fitting
for batch in doc_chunks:
    topic_model.partial_fit(batch)

topic_model.save("temp", serialization="safetensors") # -> This throws the error
```

### BERTopic Version

0.17.3

sebastianbreguel:
# What does this PR do?

Fixes #2432

`BERTopic.save(serialization=\"safetensors\")` crashes with `TypeError` after `partial_fit` discovers new clusters.

## Root cause

`TopicMapper.add_new_topics` fills the intermediate history columns of new rows with `None` placeholders:

```python
to_append = [key] + ([None] * (length - 2)) + [value]
```

The `TopicMapper` class docstring documents `mappings_` as *\"A matrix indicating the mappings from one topic to another\"* — i.e., integers — and `_save_utils.save_topics` relies on that contract:

```python
\"topic_mapper\": np.array(model.topic_mapper_.mappings_, dtype=int).tolist(),
```

`None` is not castable to `int`, so the cast raises `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'` whenever a model is saved after `partial_fit` produced new clusters.

## Fix

Backfill the intermediate columns with the topic's own `key` instead of `None`, keeping `mappings_` a homogeneous integer matrix:

```python
to_append = [key] * (length - 1) + [value]
```

The fix is at the source rather than in `_save_utils.py`: the docstring promises an integer matrix, and the `None` placeholders were a contract violation. `get_mappings` only ever reads columns `[0, -1]` (or `[-3, -1]`) and never relied on the `None` sentinel for new rows, so callers are unaffected. Pre-existing rows added by `__init__` and `add_mappings` already contained only integers, so the new rows are now consistent with them.

## Tests

Adds `test_topic_mapper_add_new_topics_keeps_integer_matrix` in `tests/test_bertopic.py`. The test:

1. Builds a `TopicMapper` and simulates two prior \`reduce_topics\` calls so the matrix has more than 2 columns (otherwise the buggy `length - 2` path is hidden).
2. Calls `add_new_topics({3: 2, 4: 3})` to mimic new clusters being discovered during `partial_fit`.
3. Asserts the matrix round-trips through `np.array(..., dtype=int)` — exactly what `_save_utils.save_topics` does.
4. Asserts pre-existing rows are untouched.
5. Asserts the original (`key`) and current (`value`) state of new rows are preserved, and the intermediate columns are backfilled with the topic's own key.

`uv run pytest tests/test_bertopic.py` → 10 passed, 1 skipped (cuML), confirming nothing downstream (including `online_topic_model`'s `partial_fit` flow) depended on the `None` sentinel. `ruff check` and `ruff format --check` are clean.

## Before submitting

### DIFF
--- a/bertopic/_bertopic.py
+++ b/bertopic/_bertopic.py
@@ -5006,12 +5006,18 @@ class TopicMapper:
     def add_new_topics(self, mappings: Mapping[int, int]):
         """Add new row(s) of topic mappings.
 
+        New topics did not exist at earlier states, so the intermediate
+        history columns are backfilled with the topic's own ``key`` to
+        keep ``mappings_`` a homogeneous integer matrix. Without this,
+        ``None`` placeholders break ``model.save(serialization="safetensors")``
+        which casts the matrix to ``np.array(..., dtype=int)``.
+
         Arguments:
             mappings: The mappings to add
         """
         length = len(self.mappings_[0])
         for key, value in mappings.items():
-            to_append = [key] + ([None] * (length - 2)) + [value]
+            to_append = [key] * (length - 1) + [value]
             self.mappings_.append(to_append)
 
 
--- a/tests/test_bertopic.py
+++ b/tests/test_bertopic.py
@@ -1,6 +1,8 @@
 import copy
 import pytest
+import numpy as np
 from bertopic import BERTopic
+from bertopic._bertopic import TopicMapper
 import importlib.util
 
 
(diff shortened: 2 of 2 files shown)


sebastianbreguel:
@MaartenGr could you review this PR?

